PR #74, opened 2026-03-22
Inject document content when using @-mentions

conbrodie:
## Summary

Adds document `@`-mentions to chat. Users can search indexed documents from the composer, attach one or more references, and send them alongside text and uploads. Mentioned document content is fetched and injected into the LLM request when the message is processed.

### Mention composer and rendering

- Adds document typeahead to the chat composer with keyboard and mouse selection.
- Requires at least 3 normalized Unicode characters before searching.
- Uses case-insensitive, punctuation-normalized prefix matching without fuzzy typo matching.
- Displays source-aware document icons with a generic fallback for historical metadata.
- Renders mention chips in sent messages.
- Preserves authorized mention/upload blocks when retrying or editing messages, including mention-only messages.

### Search and typeahead

- Maintains an in-memory document-title index refreshed hourly.
- Refreshes the title index with keyset pagination.
- Limits the ranked candidate set to 100 before ACL filtering.
- Filters unsupported content types and soft-deleted sources.

### Persistence and LLM context

- Stores compact mention references as custom document blocks in the existing `chat_messages.message` JSON; no document content is duplicated in chat persistence.
- Expands mention references per chat turn by fetching the current document content.
- Supports text and binary/PDF fallback paths through the document handler.
- Converts mention document blocks for OpenAI, OpenAI-compatible, Gemini, Anthropic, and Bedrock provider payloads.
- Leaves compaction behavior unchanged; persisted references can be expanded again on later turns.

### Permissions

Mention typeahead, expansion, and explicit document reads use aligned access rules:

- Public documents
- Direct user grants
- Domain grants
- Resolved group grants
- Soft-deleted source exclusion

Group/identity lookup failures fail closed. Permission bypass remains limited to contexts that explicitly set `skip_permission_check`.

## Validation

- Focused AI integration/unit tests: 29 passed
- Searcher typeahead unit tests: 16 passed
- `cargo check -p omni-searcher -p shared`: passed
- Searcher integration test target compilation: passed
- Changed frontend files pass Prettier

The full repository Svelte check still reports pre-existing upstream diagnostics unrelated to this PR.

## Manual testing

- [ ] Mention text-backed documents
- [ ] Mention PDF/binary documents
- [ ] Send mention-only messages
- [ ] Retry and edit messages containing mentions/uploads
- [ ] Verify public, direct-user, domain, and group access paths
- [ ] Verify unauthorized and deleted-source documents are excluded


prvnsmpth:
Awesome, thanks for the PR! Will review and leave my comments.

prvnsmpth (commented):
Overall looks great! The only major detail we should reconsider is the schema change, the rest LGTM. Thanks again!

PR #117, opened 2026-04-06
Build separate cpu-only and cuda images for docling 

prvnsmpth:
The docling workflow builds an image with gpu support by default, making it a very large image (~3 GB). Since we're including docling in our stack by default, it adds significant overhead for every install, especially because it is not enabled by default.

A better solution is to build two separate images: one meant for CPU-only usage, and one with cuda support. And our stack will use the CPU version by default, reducing the overhead significantly.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -73,6 +73,9 @@ LOCAL_EMBEDDINGS_URL=http://embeddings:${LOCAL_EMBEDDINGS_PORT}/v1
 DOCLING_URL=http://docling:${DOCLING_PORT}
 
 # Docling Configuration (AI-based document conversion)
+# DOCLING_DEVICE=cuda              # Set to "cuda" to pull the CUDA-enabled image (default: CPU-only)
+# DOCLING_CPUS=1                   # CPU limit for the docling container
+# DOCLING_MEMORY=2g                # Memory limit for the docling container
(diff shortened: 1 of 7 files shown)


prvnsmpth:
@ramteid I noticed that the docling image was ~3 GB, and it included the cuda runtime even though most people will not enable GPU support, at least not on initial set up. So I think a better solution is to create two separate images: one for cpu and one for gpu and use the cpu one by default. Users can switch if they want gpu-accelerated extraction.

ramteid:
@prvnsmpth There's a problem now when DOCLING_DEVICE is not set:
`The "DOCLING_DEVICE" variable is not set. Defaulting to a blank string. `

Also the image `ghcr.io/getomnico/omni/omni-docling` is not found when trying to pull it - was it not pushed by your pipeline?

prvnsmpth:
Oops, will fix that. The images are published, but since we don't have a proper default for DOCLING_DEVICE we don't pull the right image: https://github.com/getomnico/omni/pkgs/container/omni%2Fomni-docling

prvnsmpth:
@ramteid I've tagged the latest images with `latest` and `latest-cuda` so docker pull should work now. I've also pushed a commit with `DOCLING_DEVICE=` (empty) so that we use the cpu images by default.

PR #127, opened 2026-04-08
bugfix: Various fixes

ramteid:
This pull request introduces several improvements and new features across environment configuration, Docker Compose setup, and dependency management. The most notable changes are the addition of the Paperless connector, enhancements to Docker Compose for improved resource management and service health checks, and minor dependency and code cleanups.

**New Connector and Configuration:**

* Added `PAPERLESS_CONNECTOR_PORT` to `.env.example` and included `paperless` in the list of available connectors. [[1]](diffhunk://#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cR36) [[2]](diffhunk://#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cL53-R54)
* Added Docker Compose service definitions for `paperless-connector` in both development and production files, including build context, environment variables, and network configuration. [[1]](diffhunk://#diff-740e0cdeda40047aeb493482ba2b352edbacdc0107a84491d149bcd219dd95ffR207-R221) [[2]](diffhunk://#diff-423deb13b7c401b1a7f41ee91c77f722e11d2f317d6a66b546524e8a04cc8b03R551-R573)

**Docker Compose Improvements:**

* Updated resource limits for services using `mem_limit` and added comments for CPU pinning, replacing the previous `deploy.resources` syntax. [[1]](diffhunk://#diff-423deb13b7c401b1a7f41ee91c77f722e11d2f317d6a66b546524e8a04cc8b03L206-R207) [[2]](diffhunk://#diff-423deb13b7c401b1a7f41ee91c77f722e11d2f317d6a66b546524e8a04cc8b03L327-R325)
* Enhanced `postgres` service with a health check and updated `filesystem-connector` to depend on healthy `postgres` and completed `migrator` services. [[1]](diffhunk://#diff-740e0cdeda40047aeb493482ba2b352edbacdc0107a84491d149bcd219dd95ffR23-R24) [[2]](diffhunk://#diff-423deb13b7c401b1a7f41ee91c77f722e11d2f317d6a66b546524e8a04cc8b03L625-R652)
* Added `ORIGIN` environment variable to the `web` service for improved configuration.

**Dependency and Code Cleanups:**

* Removed `default-features = false` from `reqwest` dependencies in `connectors/slack/Cargo.toml` and `connectors/web/Cargo.toml` for simplified configuration. [[1]](diffhunk://#diff-14a28edc76d39c6d2499317d75400f392945c05b0d863fd82e0f7e9668557811L24-R24) [[2]](diffhunk://#diff-5d374c055d20180c5cb2a64c006291930c8eb8cdf6452ecfc36b04add719b50bL21-R21)
* Cleaned up unused imports in `shared/src/clients/ai.rs` and `shared/src/storage/s3.rs`. [[1]](diffhunk://#diff-942b48f1b5c3ba8d2ce46ae6aba49a99d7f5fa318aad097f46ca6f5c02fe6940L2-R6) [[2]](diffhunk://#diff-b07d1a4ad6795846fcf66783989d703ab7b72a0a6c03356af7b7c240562547c0L10)
* Removed unused `[profile.release]` section from `benchmarks/Cargo.toml`.
* Fixed an unnecessary increment in `bind_index` in `shared/src/db/repositories/embedding.rs`.

**Environment Variable Updates:**

* Updated `DOCLING_DEVICE` assignment in `.env.example` to use an explicit empty string for clarity.

These changes collectively improve the project's extensibility, maintainability, and deployment reliability.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -33,6 +33,7 @@ CLICKUP_CONNECTOR_PORT=4011
 LINEAR_CONNECTOR_PORT=4012
 FILESYSTEM_CONNECTOR_PORT=4013
 NEXTCLOUD_CONNECTOR_PORT=4014
+PAPERLESS_CONNECTOR_PORT=4015
 
 # Sandbox Port
 SANDBOX_PORT=8090
@@ -50,7 +51,7 @@ DOCLING_PORT=8003 # For AI-based document conversion via Docling
 #
 # Enable connectors you want to run by adding their profile to ENABLED_CONNECTORS (comma-separated).
 # Available connector names:
-# 	google, slack, atlassian, web, github, notion, hubspot, fireflies, microsoft, filesystem, imap, linear, clickup
+# 	google, slack, atlassian, web, github, notion, hubspot, fireflies, microsoft, filesystem, imap, linear, clickup, nextcloud, paperless
 #
 # Example: ENABLED_CONNECTORS=google,slack
 #
--- a/.env.example
+++ b/.env.example
@@ -75,9 +75,8 @@ LOCAL_EMBEDDINGS_URL=http://embeddings:${LOCAL_EMBEDDINGS_PORT}/v1
 DOCLING_URL=http://docling:${DOCLING_PORT}
 
 # Docling Configuration (AI-based document conversion)
-DOCLING_DEVICE=""   			   # Leave empty to pull the default, CPU-only image
-# DOCLING_DEVICE=cuda              # Set to "cuda" to pull the CUDA-enabled image (default: CPU-only)
-# DOCLING_CPUS=1                   # CPU limit for the docling container
+DOCLING_DEVICE=""   			   # Leave empty to pull the default, CPU-only image. Set to "cuda" to pull the CUDA-enabled image
+DOCLING_CPUSET="2,3"                 # CPU cores used for the docling container
 # DOCLING_MEMORY=2g                # Memory limit for the docling container
 DOCLING_MAX_CONCURRENT_CONVERSIONS=1
 
--- a/.env.example
+++ b/.env.example
@@ -75,7 +75,7 @@ LOCAL_EMBEDDINGS_URL=http://embeddings:${LOCAL_EMBEDDINGS_PORT}/v1
 DOCLING_URL=http://docling:${DOCLING_PORT}
 
 # Docling Configuration (AI-based document conversion)
-DOCLING_DEVICE=			   # Leave empty to pull the default, CPU-only image
+DOCLING_DEVICE=""   			   # Leave empty to pull the default, CPU-only image
 # DOCLING_DEVICE=cuda              # Set to "cuda" to pull the CUDA-enabled image (default: CPU-only)
 # DOCLING_CPUS=1                   # CPU limit for the docling container
 # DOCLING_MEMORY=2g                # Memory limit for the docling container
--- a/README.md
+++ b/README.md
@@ -36,11 +36,6 @@ The AI agent can execute code in a sandboxed container that runs on an isolated 
 
 See the full [architecture documentation](https://docs.getomni.co/architecture) for more details.
 
-## Minimum Requirements
-
-- CPU with 4 or more cores
-- 4 GB RAM
-
 ## Deployment
 
 Omni can be deployed entirely on your own infra. See our deployment guides:
--- a/README.md
+++ b/README.md
@@ -36,6 +36,11 @@ The AI agent can execute code in a sandboxed container that runs on an isolated 
 
 See the full [architecture documentation](https://docs.getomni.co/architecture) for more details.
 
+## Minimum Requirements
+
+- CPU with 4 or more cores
+- 4 GB RAM
+
 ## Deployment
 
 Omni can be deployed entirely on your own infra. See our deployment guides:
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -20,6 +20,8 @@ services:
       - "5432:5432"
     environment:
       <<: *dev-db-config
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U omni_dev -d omni_dev"]
 
   redis:
     ports:
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -202,6 +202,21 @@ services:
       context: ..
       dockerfile: connectors/microsoft/Dockerfile
 
+  filesystem-connector:
+    image: omni-filesystem-connector:dev
+    build:
+      context: ..
+      dockerfile: connectors/filesystem/Dockerfile
+    environment:
+      <<: *dev-database-url
+      RUST_BACKTRACE: full
+
+  paperless-connector:
+    image: omni-paperless-connector:dev
+    build:
+      context: ..
+      dockerfile: connectors/paperless/Dockerfile
+
   docling:
     image: omni-docling:dev
     build:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -204,7 +204,7 @@ services:
     networks:
       - sandbox-net
     mem_limit: 512m
-    # cpuset: "0"   # Uncomment to pin to specific cores, e.g. CPU=0,1
+    cpuset: "0,1"   # Pin to specific cores, e.g. CPU=0,1
     security_opt:
       - no-new-privileges:true
     read_only: true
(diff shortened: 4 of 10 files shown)


docker/docker-compose.yml
```diff
@@ -203,11 +203,8 @@ services:
       - sandbox_data:/scratch
     networks:
       - sandbox-net
-    deploy:
-      resources:
-        limits:
-          cpus: '1'
-          memory: 512M
```
ramteid, line 210: @prvnsmpth The deploy statement is not supported by the Linux kernel on various systems - e. g. my Synology NAS. There is no alternative that restricts the CPU usage by percentage. The only alternative is `cpuset` which pins the service to specific CPU cores. 

Having said that, limiting the service to one CPU does not seem to be necessary. Was that intended?

docker/docker-compose.dev.yml
```diff
@@ -20,6 +20,8 @@ services:
       - "5432:5432"
     environment:
       <<: *dev-db-config
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U omni_dev -d omni_dev"]
```
ramteid, line 24: This didn't yet use the overrides and caused issues using the dev config

docker/docker-compose.yml
```diff
@@ -649,6 +670,7 @@ services:
       SESSION_COOKIE_NAME: ${SESSION_COOKIE_NAME}
       SESSION_DURATION_DAYS: ${SESSION_DURATION_DAYS}
       APP_URL: ${APP_URL}
+      ORIGIN: ${APP_URL}
```
ramteid, line 673: Without this, the initial signup fails with an CORS error

shared/src/db/repositories/embedding.rs
```diff
@@ -175,7 +175,6 @@ impl EmbeddingRepository {
         if let Some(ct) = content_types {
             if !ct.is_empty() {
                 where_conditions.push(format!("d.content_type = ANY(${})", bind_index));
-                bind_index += 1;
```
ramteid, line 178: bind_index is not read anymore after that

.env.example
```diff
@@ -74,7 +75,7 @@ LOCAL_EMBEDDINGS_URL=http://embeddings:${LOCAL_EMBEDDINGS_PORT}/v1
 DOCLING_URL=http://docling:${DOCLING_PORT}
 
 # Docling Configuration (AI-based document conversion)
-DOCLING_DEVICE=			   # Leave empty to pull the default, CPU-only image
+DOCLING_DEVICE=""   			   # Leave empty to pull the default, CPU-only image
```
ramteid: Before, the comment could have been parsed

docker/docker-compose.yml
```diff
@@ -203,11 +203,8 @@ services:
       - sandbox_data:/scratch
     networks:
       - sandbox-net
-    deploy:
-      resources:
-        limits:
-          cpus: '1'
-          memory: 512M
+    mem_limit: 512m
```
prvnsmpth, line 206: I think `mem_limit` is deprecated, we should use `deploy.resources.limits` to specify the limits: https://docs.docker.com/reference/compose-file/deploy/#resources

And we should definitely limit CPUs to the sandbox to prevent any chances of it running something that will affect other services.

docker/docker-compose.yml
```diff
@@ -324,11 +321,8 @@ services:
     volumes:
       - docling_models:/root/.cache
     restart: unless-stopped
-    deploy:
-      resources:
-        limits:
-          cpus: "${DOCLING_CPUS:-1}"
-          memory: "${DOCLING_MEMORY:-2g}"
+    mem_limit: ${DOCLING_MEMORY:-2g}
```
prvnsmpth, line 324: Same as above, let's keep `deploy.resources.limits`

prvnsmpth (commented):
Thanks for the fixes! Just one comment about the docker mem limit

docker/docker-compose.yml
ramteid: @prvnsmpth The deploy statement simply does not work on many systems. Few people will deploy Omni on their workstation or their development notebook. Omni will run on NAS systems, Raspberry Pi or similar platforms.

Using the deploy statement with my Synology NAS I get this error:

`Error response from daemon: NanoCPUs can not be set, as your kernel does not support CPU CFS scheduler or the cgroup is not mounted`

I did a little research and found that it's a general problem in many systems. GPU throttling via CFS scheduler is not allowed.

I understand the motivation why you want to use it but it is simply not feasible in general in my opinion. If you want to keep it, you will exclude many users because Omni simply won't work. I can remove the statement manually on my systems but many users won't find the cause for this error.

If you really want to limit specific services to single CPUS, you can you can still use
`cpuset: "0"`
which will limit the service to CPU core 0. Alternate between 0 and 1 for multiple services.

I can uncomment cpuset if you agree.

ramteid:
Not sure why some checks fail?

prvnsmpth:
Let me take a look. Something is cancelling the workflow, I see this in the logs:

`Canceling since a higher priority waiting request for build-omni-atlassian-connector-refs/pull/127/merge exists`

This usually means a new commit was pushed to the same PR, but looks like the latest run failed for some reason even though no new commit was pushed. I'll look into this.

prvnsmpth:
Seems like there's some race condition with the PR invalidation logic. We cancel the previous run when a new commit is pushed to the PR. But due to this race condition, both previous and current end up cancelled.

README.md
```diff
@@ -36,6 +36,11 @@ The AI agent can execute code in a sandboxed container that runs on an isolated
 
 See the full [architecture documentation](https://docs.getomni.co/architecture) for more details.
 
+## Minimum Requirements
```
prvnsmpth: In our docs: https://docs.getomni.co/deployment/docker-compose I've listed 4 cores, 8 GB as recommended requirements. Do you think we should reduce the memory recommendation?

prvnsmpth:
@ramteid Looks good! Just one comment about the `mem_limit` flag, I think we should use the `deploy.resources` field instead.

ramteid:
> @ramteid Looks good! Just one comment about the `mem_limit` flag, I think we should use the `deploy.resources` field instead.

I disagree, as deploy.resources does not work on many systems. Check my previous comment.
Anyway, I'll revert that change and open an issue so it can be discussed there.

I will have to manually remove the deploy.resources statement on my system just to make omni work at all, that's really ugly :-/

prvnsmpth:
> I will have to manually remove the deploy.resources statement on my system just to make omni work at all, that's really ugly 

Oh! That's bad... what docker compose version are you using? Could you share the error message you see, if any?

README.md
ramteid: No, my mistake. Didn't see it was already documented somewhere. Reverted.

ramteid:
> > I will have to manually remove the deploy.resources statement on my system just to make omni work at all, that's really ugly
> 
> Oh! That's bad... what docker compose version are you using? Could you share the error message you see, if any?

It's not a docker compose problem - it's a problem when the host Linux kernel does not enable CFS - what is apparently pretty common.
https://github.com/getomnico/omni/pull/127#discussion_r3056464895

docker/docker-compose.yml
prvnsmpth: Oh I see, that makes sense. We can use `mem_limit` and `cpus` in that case.

Yeah I had deliberately limited cpu to 1 for the sandbox because I don't want to leave any opportunities for arbitrary code executed by the models to affect other services too much. The code generated by models will only be used for analyzing files, writing one-off python scripts, etc. so it makes sense to keep the resource utilization to a minimum.

docker/docker-compose.yml
prvnsmpth: Yes I had noticed this too but for some reason, all my deployments are running without this :man_shrugging: 

But yeah this is the correct way to explicitly set the origin for CORS in sveltekit.

prvnsmpth:
> It's not a docker compose problem - it's a problem when the host Linux kernel does not enable CFS - what is apparently pretty common. [#127 (comment)](https://github.com/getomnico/omni/pull/127#discussion_r3056464895)

Oh I see, let's use `mem_limit` in that case. We can make the change in this PR, thanks! 

edit: and about the `cpuset`, we can use "0,1" I think, that shouldn't be a problem.

prvnsmpth:
BTW, I had never even considered NASs and Raspberry Pi as viable deployment targets for the project :) Great to see Omni actually being tested on such varied hardware!

ramteid:
> > It's not a docker compose problem - it's a problem when the host Linux kernel does not enable CFS - what is apparently pretty common. [#127 (comment)](https://github.com/getomnico/omni/pull/127#discussion_r3056464895)
> 
> Oh I see, let's use `mem_limit` in that case. We can make the change in this PR, thanks!
> 
> edit: and about the `cpuset`, we can use "0,1" I think, that shouldn't be a problem.

Alright, I will re-apply the change from deploy to cpuset and mem_limit. Please check. Can be merged if there are no other concerns :)

PR #135, opened 2026-04-09
feat: Telegram connector with Telethon + Bot API backends and admin UI

xingfanxia:
## Summary

Adds a **Telegram connector** to Omni with full admin UI support. Two backends:

- **Telethon (MTProto)** — full chat history via user session string. Requires `api_id`, `api_hash`, `session`.
- **Bot API** — forward-only message capture via bot token. Requires `token`.

The backend is auto-detected from the credentials shape. An interactive auth script (`connectors/telegram/scripts/auth.py`) walks users through a one-time phone login to generate a Telethon session string.

Indexes in production at https://omni.computelabs.ai — currently syncing ~17K Telegram messages across two sources (admin personal + team BD account, 98 chats total).

## What's included

**Backend (connector + docker + enums)**:
- `connectors/telegram/` — Python connector with Telethon + Bot API backends, Dockerfile, auth script, README
- `docker/docker-compose.yml` — `telegram-connector` service under the `telegram` profile
- `shared/src/models.rs` — `Telegram` variant on `SourceType` and `ServiceProvider`
- `web/src/lib/types.ts` — matching TypeScript enum entries

**Features**:
- Chat discovery via `list_chats` connector action (Telethon only)
- Selective sync by chat title (`chats`) or chat ID (`chat_ids`)
- Incremental sync via per-chat `last_message_id` checkpointing
- **Per-source admin restriction via optional `allowed_users` config**: when set, emitted documents get `permissions.public=false, users=[...]`. Useful for sensitive sources (e.g. BD / investor conversations) where only specific users should be able to search the content.

**Admin UI**:
- Setup dialog with Telethon / Bot Token mode switcher, linked setup instructions for `my.telegram.org/apps` and BotFather
- Settings page with **Load Chats** button (calls `list_chats` action), chat picker with type badge + member count, and cross-source dedup indicator that surfaces already-synced chats when adding a second Telegram source
- **Add Another Account** button on the settings page to open the setup dialog in-place (supports multi-account setups without leaving the settings context)
- Telegram integration card on the admin integrations page

**Bug fixes uncovered during integration**:
- `execute_action` now unwraps the credential envelope: the connector-manager wraps credentials as `{credentials, config, principal_email}` for action dispatch but passes them raw for sync. The Python connector's `_is_telethon()` was checking top-level keys, so the action path always failed. Fix in `telegram_connector/connector.py`.
- Integer type coercion for `chat_id` / `message_id` (JSON deserialization yields strings in some paths)
- ISO string / datetime / unix timestamp parsing in `mappers.py:_parse_date()`

## Test plan

- [x] Telethon sync against real account — 7,743 + 9,093 messages indexed across 98 chats (two sources)
- [x] `list_chats` action returns 1,600+ dialogs for a user account
- [x] Incremental re-sync: no-op run on an up-to-date source completes in <5s with 0 new messages
- [x] Web UI end-to-end: setup dialog → source created → Load Chats → pick chats → save → sync triggered → messages searchable via `/api/v1/search`
- [x] `allowed_users`: verified docs emitted with `permissions.users=[email]` instead of `public=true`; restricted API keys return zero telegram results
- [x] Web build includes all new routes (`grep 'integrations/telegram' build/server/manifest.js` matches)
- [ ] Automated tests — none yet for this connector (happy to add basic ones if desired; following the pattern of other connectors which also ship without tests)

## Screenshots

(Happy to add screenshots of the setup dialog and chat picker if useful for review.)

## Notes for reviewers

- Uses Telethon's `StringSession` — note that session strings are long-lived credentials equivalent to full account access. The README documents rotation and security considerations.
- The `allowed_users` filter is enforced at document creation time (via `permissions.users` JSONB). It integrates with the existing searcher permission filter without any Rust changes.
- No database migrations required beyond adding `'telegram'` to the existing `sources.source_type` and `service_credentials.provider` CHECK constraints (if those constraints exist in your deployment — ours did, added via an earlier local migration). If you prefer, happy to include a migration file.


connectors/telegram/scripts/auth.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
```
prvnsmpth, line 1: Great work!

Is there any easy way to incorporate this setup step in the UI? Or maybe we can do that later as a follow-up?

xingfanxia: Thanks! Yes, doable — happy to do it as a follow-up PR right after this one merges to keep this PR focused.

The rough plan:

**New Python connector actions** (stateless, frontend holds transient state between steps):

1. `auth_send_code`
   - Input: `{api_id, api_hash, phone}`
   - Calls `client.send_code_request(phone)`
   - Returns `{partial_session, phone_code_hash}` (partial StringSession has the DC auth key but no logged-in user yet — not sensitive on its own, expires after a few minutes on Telegram's side)

2. `auth_verify_code`
   - Input: `{api_id, api_hash, partial_session, phone, phone_code_hash, code, password?}`
   - Reconstructs the client from `StringSession(partial_session)` and calls `sign_in(phone, code, phone_code_hash=...)`
   - If Telethon raises `SessionPasswordNeededError` → return `{needs_2fa: true, partial_session}` and the frontend shows a password input
   - On success → returns `{session}` (full StringSession the frontend writes into the source credentials)

**UI** (new component in `web/src/lib/components/telegram-auth-flow.svelte`, wired into the existing setup dialog):

```
┌─ Step 1: API credentials + phone ─┐
│ api_id:    [_______________]      │
│ api_hash:  [_______________]      │
│ phone:     [+1 555 123 4567]      │  → "Send code" button
└────────────────────────────────────┘
         ↓ (calls auth_send_code)
┌─ Step 2: SMS code ────────────────┐
│ code:  [_____]                     │  → "Verify" button
└────────────────────────────────────┘
         ↓ (calls auth_verify_code)
  if needs_2fa:
┌─ Step 3: 2FA password ────────────┐
│ password: [_______________]        │  → "Submit" button
└────────────────────────────────────┘
         ↓ (calls auth_verify_code again with password)
  → session string written directly into source credentials, no copy-paste

```

The existing "paste pre-generated session string" path in the setup dialog stays as a fallback for users who want to use `scripts/auth.py` anyway (CI, air-gapped runs, etc.).

I'll open a follow-up PR as soon as this one lands.

xingfanxia: Addressed — just pushed `8ec0ef4`. The setup step is now fully in the UI. Summary of what changed (same shape as my earlier plan, implemented end-to-end):

**Python connector** — two new stateless actions in `telegram_connector.connector`:

- `auth_send_code({api_id, api_hash, phone})` calls `TelegramClient.send_code_request` on a fresh `StringSession()`, serializes the in-flight session, and returns `{partial_session, phone_code_hash}`. The partial session carries only the DC auth key (no user logged in yet), so it's safe to hand back to the frontend between steps.
- `auth_verify_code({api_id, api_hash, partial_session, phone, phone_code_hash, code, password?})` reconstructs the client from the partial session and calls `sign_in`. On `SessionPasswordNeededError` with no `password`, returns `{needs_2fa: true, partial_session}` — reusing the same partial session avoids burning a fresh SMS code. On success returns `{session, user: {...}}` so the UI can show "Signed in as …".

Neither action needs pre-existing credentials — all input comes from `params` — so the flow runs before any source exists.

**Web backend** — new `POST /api/connectors/telegram/auth` route that:

- Is admin-only.
- Whitelists only `auth_send_code` / `auth_verify_code` to keep the source-less bypass narrow.
- Discovers the telegram connector's URL via `${connectorManagerUrl}/connectors` (no extra env var needed).
- POSTs directly to the Python connector's `/action` endpoint with an empty credentials envelope.

**Web UI** — `telegram-connector-setup.svelte` gains a sub-method toggle in Telethon mode:

- **Sign in with phone number (recommended)** — new default. Wizard states: `start → code → (twofa) → done`. Collects phone → calls `auth_send_code` → collects SMS code → calls `auth_verify_code` → if `needs_2fa`, collects password and calls again with the same partial session / code / hash → on `done`, shows a green "Signed in as …" banner and unblocks the outer **Connect** button.
- **Paste existing session string** — kept as a fallback for CI / air-gapped / power-user scenarios. `scripts/auth.py` stays in the repo for this path.

**Smoke-tested** end-to-end on my dev instance. With deliberately-invalid api_id/hash, `POST /api/connectors/telegram/auth` returns

```json
{"ok": false, "error": "Failed to send code: The api_id/api_hash combination is invalid (caused by SendCodeRequest)"}
```

— which confirms the full path (web → connector-manager discovery → telegram-connector → Telethon → Telegram's API) is wired correctly. The error surfaces back to the UI's `authError` banner.

README updated to present the UI as the recommended path and `scripts/auth.py` as the fallback.

web/src/routes/api/connectors/telegram/auth/+server.ts
```diff
@@ -0,0 +1,119 @@
...
+import { logger } from '$lib/server/logger'
+
+// Pre-source auth action dispatch for the Telegram connector.
+//
+// Unlike `/api/sources/:id/action` (which looks up credentials from an
+// existing source), this route runs auth actions BEFORE a source exists —
+// the whole point is to help the user build their initial credentials via
+// the web UI instead of `scripts/auth.py`.
+//
+// Strategy: ask connector-manager which URL the telegram connector is
```
prvnsmpth: Hmm, this kinda breaks the abstraction boundary we have around connectors - all interactions between omni services and connectors go via the connector mgr (sdk client for calls from connector -> omni, and the /actions endpoint for the other direction). Would be nice if we can keep this abstraction.

I understand that for auth we need to make an unauthenticated call to the connector. To support this can we make the credential load optional for certain actions? Maybe we can have an optional field defined on the action in the manifest, something like `authenticated: false` and that will mean that the action doesn't require auth. WDYT?

xingfanxia: Good point — agree we should keep the abstraction boundary intact.

Here's my proposed approach:

**1. Connector SDK** — add an `authenticated: bool` (default `true`) field to `ActionDefinition`. The telegram connector marks `auth_send_code` and `auth_verify_code` as `authenticated: false`.

**2. Connector-manager** — add a source-type-scoped action endpoint (e.g., `POST /connectors/:source_type/action`) that:
- Looks up the connector URL from the registry (same data as `/connectors`)
- Fetches the action manifest from the connector and checks `authenticated`
- If `authenticated: false` → forwards to the connector's `/action` with empty credentials
- If `authenticated: true` → rejects with 400 ("this action requires a source ID")

This keeps the routing through connector-manager while carving a narrow exception for pre-auth flows. The pattern is also reusable for any future connector that needs pre-source actions (e.g., OAuth discovery, workspace auth).

**3. Web route** — simplify `telegram/auth/+server.ts` to call connector-manager's new endpoint instead of doing its own connector URL discovery. The admin-only guard and action whitelist stay in the web layer as defense-in-depth.

The existing `POST /api/sources/:id/action` path (which requires a source ID for credential loading) stays untouched.

Will rework and push. Let me know if this shape looks right or if you'd prefer a different approach.

PR #139, opened 2026-04-10
Fix: test env refine

cuberone:
## Summary

**Linux-only Landlock sandboxing** — the `landlock` crate in `sandbox_exec.rs` are now gated behind `#[cfg(target_os = "linux")]`, fixing cross-platform build failures. The sandbox binary now compiles on macOS/Windows while Linux deployments retain full filesystem isolation. Also adds three new required fields (`mcp_enabled`, `prompts`, `resources`) to fix some failed tests.

**MinIO for S3 integration tests** — added lightweight (compared to LocalStack) MinIO container in `TestEnvironment` as some tests demand s3 running.

**Fix `.env.example` inline comments** — comments on the same line as `DOCLING_DEVICE=` were being parsed as part of the value by some shell parsers. Moved to separate lines.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -75,13 +75,10 @@ LOCAL_EMBEDDINGS_URL=http://embeddings:${LOCAL_EMBEDDINGS_PORT}/v1
 DOCLING_URL=http://docling:${DOCLING_PORT}
 
 # Docling Configuration (AI-based document conversion)
-DOCLING_DEVICE=""   			   # Leave empty to pull the default, CPU-only image. Set to "cuda" to pull the CUDA-enabled image
-DOCLING_CPUSET="2,3"                 # CPU cores used for the docling container
-# DOCLING_MEMORY=2g                # Memory limit for the docling container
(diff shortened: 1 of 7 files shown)


prvnsmpth:
Thanks for the PR! I'd been meaning to fix those s3-related test failures for a while, thanks for fixing them :)

ramteid:
Minio is deprecated. It has become commercial and the open source version has been abandoned. Shouldn't use components that are no longer maintained due to security risks.

cuberone:
Switched to localstack

cuberone:
`lto` = `"thin"` seems to have been unintentionally introduced in this PR.
I wasn’t able to build locally due to OOM on a 16GB Mac without this optimization flag.
I’d prefer to keep "thin" enabled, as the performance impact should be minimal.

cuberone:
I've just found that localstack was archived recently as well) But both should work as test stub even Minio being AGPL3

prvnsmpth:
@cuberone Thanks! We can use localstack, shouldn't be a big issue since it's a dev dependency. We can switch later if we find a better alternative.

PR #144, opened 2026-04-12
fix: allow multiple openai_compatible providers and reduce AI dev container CPU usage

cuberone:
## Summary

- **Allow multiple OpenAI-compatible providers** — the LLM settings page previously hid `openai_compatible` from the "Connect a Provider" grid after the first instance was configured (treating it like a singleton). Introduces `MULTI_INSTANCE_TYPES` so it always remains available, and rewrites `connectedProviders` to iterate `data.providers` directly, rendering each configured instance as an independent card. The motivating use case is running OpenRouter and a local LLM simultaneously — both as separate `openai_compatible` entries.
- **Reduce AI service dev container CPU usage** — adds an anonymous volume at `/app/.venv` in `docker-compose.dev.yml` to mask the host `.venv` directory from the container. No impact on hot reload; eliminates unnecessary inotify watching of the virtualenv, cutting CPU usage by up to 30%.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -87,6 +87,7 @@ services:
       LOG_LEVEL: debug
     volumes:
       - ../services/ai:/app
+      - /app/.venv
     command: python -m uvicorn main:app --host 0.0.0.0 --port ${AI_SERVICE_PORT} --reload
 
   sandbox:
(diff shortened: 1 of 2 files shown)


prvnsmpth:
Thanks for the PR! Merged.

PR #146, opened 2026-04-12
Address docling memory usage issues

prvnsmpth:
- docling currently accepts all incoming convert requests, and this leads to high memory use, stalling the conversion
- we add a limit to the number of pending requests - return 429 so that the source can back-off and slow down
- add UI controls to allow users to tradeoff quality vs speed in the conversion process, on the fly

### DIFF
--- a/sdk/python/omni_connector/storage.py
+++ b/sdk/python/omni_connector/storage.py
@@ -1,3 +1,4 @@
+import asyncio
 import base64
 import logging
 from typing import TYPE_CHECKING
@@ -5,7 +6,11 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from .client import SdkClient
 
+from .exceptions import ServiceOverloadedError
+
 logger = logging.getLogger(__name__)
+
+_MAX_RETRIES = 5
 
 
 class ContentStorage:
@@ -39,13 +44,29 @@ class ContentStorage:
         """Extract text from binary file content and store it, returning content_id.
 
         The connector manager handles extraction based on MIME type.
+        Retries with exponential backoff when the extraction service is overloaded.
         """
-        return await self._client.extract_and_store_content(
-            self._sync_run_id,
-            data,
-            mime_type,
-            filename,
-        )
+        for attempt in range(_MAX_RETRIES):
+            try:
+                return await self._client.extract_and_store_content(
+                    self._sync_run_id,
+                    data,
+                    mime_type,
+                    filename,
+                )
+            except ServiceOverloadedError as e:
+                if attempt == _MAX_RETRIES - 1:
+                    raise
+                wait = e.retry_after * (1.5**attempt)
+                logger.warning(
+                    "Extraction service overloaded for file %s, retrying in %.0fs (%d/%d)",
+                    filename,
+                    wait,
+                    attempt + 1,
+                    _MAX_RETRIES,
+                )
+                await asyncio.sleep(wait)
+        raise RuntimeError("unreachable")
 
     async def extract_text(
         self,
@@ -56,13 +77,29 @@ class ContentStorage:
         """Extract text from binary file content without storing.
 
         Use when you need to post-process or combine text before storing.
+        Retries with exponential backoff when the extraction service is overloaded.
         """
-        return await self._client.extract_text(
-            self._sync_run_id,
-            data,
-            mime_type,
-            filename,
-        )
+        for attempt in range(_MAX_RETRIES):
+            try:
+                return await self._client.extract_text(
+                    self._sync_run_id,
+                    data,
+                    mime_type,
+                    filename,
+                )
+            except ServiceOverloadedError as e:
+                if attempt == _MAX_RETRIES - 1:
+                    raise
+                wait = e.retry_after * (1.5**attempt)
+                logger.warning(
+                    "Extraction service overloaded for file %s, retrying in %.0fs (%d/%d)",
+                    filename,
+                    wait,
+                    attempt + 1,
+                    _MAX_RETRIES,
+                )
+                await asyncio.sleep(wait)
+        raise RuntimeError("unreachable")
 
     async def save_binary(
         self,
--- a/services/connector-manager/src/handlers.rs
+++ b/services/connector-manager/src/handlers.rs
@@ -19,6 +19,7 @@ use futures::stream::Stream;
 use redis::AsyncCommands;
 use serde_json::json;
 use shared::clients::docling::{DoclingClient, DoclingError};
+use shared::constants::REDIS_SYSTEM_SETTINGS_KEY;
 use shared::db::repositories::SyncRunRepository;
 use shared::models::{ConnectorManifest, SearchOperator, SourceType, SyncType};
 use shared::queue::EventQueue;
@@ -833,8 +834,29 @@ async fn is_docling_enabled(redis_client: &redis::Client) -> bool {
         }
     };
 
-    let value: Option<String> = conn.hget("system:settings", "docling_enabled").await.ok();
+    let value: Option<String> = conn
+        .hget(REDIS_SYSTEM_SETTINGS_KEY, "docling_enabled")
+        .await
+        .ok();
     value.as_deref() == Some("true")
+}
+
+/// Read the Docling quality preset from Redis (set via the admin UI).
+/// Defaults to "balanced" if not set or on error.
+async fn get_docling_quality_preset(redis_client: &redis::Client) -> String {
+    let mut conn = match redis_client.get_multiplexed_async_connection().await {
+        Ok(c) => c,
+        Err(e) => {
+            warn!("Failed to connect to Redis for docling preset: {}", e);
+            return "balanced".to_string();
+        }
+    };
+
+    let value: Option<String> = conn
+        .hget(REDIS_SYSTEM_SETTINGS_KEY, "docling_quality_preset")
+        .await
+        .ok();
+    value.unwrap_or_else(|| "balanced".to_string())
 }
 
 /// MIME types that Docling can process.
@@ -1040,11 +1062,12 @@ async fn do_extract_text(
     if docling_candidate && docling_enabled {
         let docling_result = if let Some(client) = DoclingClient::from_env() {
             let file_name = filename.unwrap_or("document");
+            let preset = get_docling_quality_preset(redis_client).await;
             debug!(
-                "Using docling-based document content extraction for file '{}'",
-                file_name
+                "Using docling-based document content extraction for file '{}' (preset={})",
+                file_name, preset
             );
-            match client.convert(data, file_name).await {
+            match client.convert(data, file_name, &preset).await {
                 Ok(markdown) => {
                     debug!("Docling extraction succeeded: {} chars", markdown.len());
                     Some(markdown)
--- a/services/connector-manager/src/handlers.rs
+++ b/services/connector-manager/src/handlers.rs
@@ -18,7 +18,7 @@ use axum::{
 use futures::stream::Stream;
 use redis::AsyncCommands;
 use serde_json::json;
-use shared::clients::docling::DoclingClient;
+use shared::clients::docling::{DoclingClient, DoclingError};
 use shared::db::repositories::SyncRunRepository;
 use shared::models::{ConnectorManifest, SearchOperator, SourceType, SyncType};
 use shared::queue::EventQueue;
@@ -653,6 +653,9 @@ pub enum ApiError {
 
     #[error("Internal error: {0}")]
     Internal(String),
+
+    #[error("Too many requests: {0}")]
+    TooManyRequests(String),
 }
 
 impl From<SyncError> for ApiError {
@@ -692,6 +695,7 @@ impl IntoResponse for ApiError {
             ApiError::BadRequest(msg) => (StatusCode::BAD_REQUEST, msg.clone()),
             ApiError::Conflict(msg) => (StatusCode::CONFLICT, msg.clone()),
             ApiError::Internal(msg) => (StatusCode::INTERNAL_SERVER_ERROR, msg.clone()),
+            ApiError::TooManyRequests(msg) => (StatusCode::TOO_MANY_REQUESTS, msg.clone()),
         };
 
         let body = json!({ "error": message });
@@ -918,10 +922,10 @@ fn is_docling_supported_extension(filename: Option<&str>) -> bool {
 
 use crate::models::{
     SdkCancelSyncRequest, SdkCancelSyncResponse, SdkCompleteRequest, SdkCreateSyncRequest,
-    SdkCreateSyncResponse, SdkEmitEventRequest, SdkExtractContentResponse,
-    SdkExtractTextResponse, SdkFailRequest, SdkIncrementScannedRequest,
-    SdkSourceSyncConfigResponse, SdkStatusResponse, SdkStoreContentRequest,
-    SdkStoreContentResponse, SdkUserEmailResponse, SdkWebhookNotification, SdkWebhookResponse,
+    SdkCreateSyncResponse, SdkEmitEventRequest, SdkExtractContentResponse, SdkExtractTextResponse,
+    SdkFailRequest, SdkIncrementScannedRequest, SdkSourceSyncConfigResponse, SdkStatusResponse,
+    SdkStoreContentRequest, SdkStoreContentResponse, SdkUserEmailResponse, SdkWebhookNotification,
+    SdkWebhookResponse,
 };
 
 pub async fn sdk_emit_event(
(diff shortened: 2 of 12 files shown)


prvnsmpth:
@ramteid Would appreciate a review on this change when you get the chance.

prvnsmpth:
image

The new document conversion settings UI.

ramteid (commented):
Nice work overall! The backpressure mechanism (429 + retry) and chunked PDF processing are solid improvements. A few items to address below.

services/connector-manager/src/handlers.rs
```diff
@@ -692,6 +696,7 @@ impl IntoResponse for ApiError {
             ApiError::BadRequest(msg) => (StatusCode::BAD_REQUEST, msg.clone()),
             ApiError::Conflict(msg) => (StatusCode::CONFLICT, msg.clone()),
             ApiError::Internal(msg) => (StatusCode::INTERNAL_SERVER_ERROR, msg.clone()),
+            ApiError::TooManyRequests(msg) => (StatusCode::TOO_MANY_REQUESTS, msg.clone()),
```
ramteid: **Bug: `Retry-After` header is lost here.** The Python SDK reads `response.headers.get("retry-after", "30")` to set the backoff delay, but this `IntoResponse` impl only returns a status + JSON body — no `Retry-After` header is emitted.

The `retry_after_secs` value from Docling is captured at line 1075 but discarded when converted to `ApiError::TooManyRequests(String)`. The SDK silently falls back to a fixed 30s.

Suggestion: either carry the retry-after value in the enum variant (`TooManyRequests { message: String, retry_after: u64 }`) and emit the header here, or document that the SDK always uses 30s.

shared/src/clients/docling.rs
```diff
@@ -153,35 +177,51 @@ impl DoclingClient {
...
-            .mime_str("application/octet-stream")?;
+            .mime_str("application/octet-stream")
+            .context("Failed to build multipart request")?;
 
         let form = multipart::Form::new().part("file", part);
 
         let response = self
             .client
-            .post(format!("{}/convert", self.base_url))
+            .post(format!("{}/convert?preset={}", self.base_url, preset))
```
ramteid: **Nit: prefer `reqwest` query builder over string interpolation.** Current presets are safe ASCII, but using `.query(&[("preset", preset)])` is more robust against unexpected values and avoids potential URL encoding issues.

services/docling/app.py
```diff
@@ -1,309 +1,463 @@
...
+        logger.info(
+            "Job %s: conversion started (%s, %d bytes, preset=%s).",
+            job.id,
+            filename,
+            len(data),
+            job.preset,
+        )
+        t0 = time.monotonic()
+        loop = asyncio.get_running_loop()
+        converter = _converters[job.preset]
```
ramteid: **Thread safety concern.** `_converters` holds one `DocumentConverter` per preset, but the semaphore only limits total concurrency — not per-preset. With `_MAX_CONCURRENT > 1`, two jobs with the same preset can hit the same converter instance concurrently via `run_in_executor` (thread pool).

The old code used a `Queue` of converter instances to guarantee exclusive access. If `DocumentConverter.convert()` is not thread-safe, this can cause data races. Worth verifying, or using a per-preset lock.

services/docling/app.py
```diff
@@ -1,309 +1,463 @@
...
+        "images_scale": 1.0,
+        "do_picture_classification": True,
+        "generate_picture_images": False,
+        "generate_table_images": False,
+    },
+    "quality": {
+        "table_structure_mode": TableFormerMode.ACCURATE,
+        "images_scale": 1.5,
+        "do_picture_classification": True,
+        "generate_picture_images": True,
```
ramteid: `generate_picture_images` and `generate_table_images` tell Docling to crop and retain pixel data for detected regions on the `ConversionResult`. Since the code only calls `result.document.export_to_markdown()` and discards the result, these image bytes are never used — they just consume extra memory during conversion.

The settings that affect text quality are `images_scale`, `do_picture_classification`, and `table_structure_mode`. These two should be `False` here, same as the other presets.

services/docling/app.py
```diff
@@ -1,309 +1,463 @@
...
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+_MAX_CONCURRENT = int(os.getenv("MAX_CONCURRENT_CONVERSIONS", "1"))
+_MAX_PENDING = int(os.getenv("MAX_PENDING_JOBS", str(_MAX_CONCURRENT * 2)))
+_PAGES_PER_CHUNK = int(os.getenv("PAGES_PER_CHUNK", "5"))
+_ready: bool = False
+
+PresetName = Literal["fast", "balanced", "quality"]
```
ramteid: **Nit:** `PresetName` is defined but never referenced. Either use it to type-annotate the preset parameters/dict keys, or remove it.

sdk/python/omni_connector/storage.py
```diff
@@ -39,13 +44,29 @@ async def extract_and_store_content(
...
         The connector manager handles extraction based on MIME type.
+        Retries with exponential backoff when the extraction service is overloaded.
         """
-        return await self._client.extract_and_store_content(
-            self._sync_run_id,
-            data,
-            mime_type,
-            filename,
-        )
+        for attempt in range(_MAX_RETRIES):
```
ramteid, line 49: **Nit: duplicated retry logic.** This retry-with-backoff block is copy-pasted identically in `extract_text` below (line 82). Consider extracting a small helper. The `raise RuntimeError("unreachable")` sentinel at the end of each is a code smell from the `for`-loop pattern.

services/connector-manager/src/handlers.rs
```diff
@@ -1029,18 +1055,32 @@ async fn do_extract_text(
     mime_type: &str,
     filename: Option<&str>,
     data: &[u8],
-) -> String {
+) -> Result<String, ApiError> {
+    let docling_enabled = is_docling_enabled(redis_client).await;
```
ramteid: **Minor:** `is_docling_enabled` and `get_docling_quality_preset` each open a separate Redis connection. Since they are always called together here, they could share a single multiplexed connection.

web/src/routes/(admin)/admin/settings/document-conversion/+page.svelte
```diff
@@ -21,91 +50,155 @@
...
                         }}>
-                        <Switch
-                            name="enabled"
-                            value="true"
-                            checked={doclingEnabled}
-                            disabled={isSubmitting}
-                            onCheckedChange={(checked) => {
-                                doclingEnabled = checked
-                                formRef?.requestSubmit()
+                        <input type="hidden" name="preset" value={qualityPreset} />
```
ramteid: **Nit:** This hidden input is redundant — the `enhance` callback already does `formData.set("preset", qualityPreset)` a few lines above. You can remove this `<input>`.

services/docling/app.py
ramteid: **Revised proposal for quality presets (verified against Docling v2.86 docs):**

| Parameter | Fast | Balanced | Quality |
|---|---|---|---|
| `table_structure_mode` | FAST | ACCURATE | ACCURATE |
| `images_scale` | 1.0 | 1.0 | 1.5 |
| `do_picture_classification` | False | True | True |
| `generate_picture_images` | False | False | **False** |
| `generate_table_images` | False | False | **False** |
| `do_code_enrichment` | False | False | **True** |
| `do_formula_enrichment` | False | False | **True** |

### Verification results

**Model requirements for code/formula enrichment:**
- Both `do_code_enrichment` and `do_formula_enrichment` use the **Code & Formula Stage** which defaults to the `codeformulav2` preset — a specialized ~0.2B SAM-OPT model ([docling-project/CodeFormula](https://huggingface.co/ds4sd/CodeFormula)), _not_ a large VLM like Granite Vision.
- Alternative preset: `granite_docling` (Granite-Docling-258M) — also small.
- No Granite Vision or large VLM download required.

**Compatibility:**
- ✅ Fully compatible with RapidOCR, table structure, picture classification, and all other settings in the pipeline.
- ✅ These are standard `PdfPipelineOptions` fields — no pipeline type change needed.
- ✅ Code enrichment operates on `CodeItem` elements, formula enrichment on `TextItem` with FORMULA label — independent of other enrichment stages.
- ✅ No known incompatibilities between any of the proposed settings.

### Approximate RAM / VRAM consumption

| Feature | Fixed cost (model load) | Per-document cost | Notes |
|---|---|---|---|
| Layout (Heron) | ~150–200 MB | negligible | Always loaded |
| TableFormer (accurate) | ~100–200 MB | negligible | Always loaded when `do_table_structure=True` |
| RapidOCR (torch) | ~50–100 MB | negligible | Always loaded when `do_ocr=True` |
| **Picture classifier** (EfficientNet-B0) | **~16 MB** | negligible | 4.08M params F32. Tiny. |
| **CodeFormulaV2** | **~900 MB** | ~per code/formula element | 0.2B params F32, `model.safetensors` = 894 MB. Loaded once, shared across jobs. |
| **`generate_picture_images: True`** | 0 | **Unbounded — scales with doc** | Stores extracted PIL images in memory. ~3–10 MB per picture at `images_scale=1.5`. A 50-page image-heavy PDF can easily add 200+ MB per conversion. |
| **`generate_table_images: True`** | 0 | **Unbounded — scales with doc** | Same: stores table screenshots as PIL images. |

**Memory comparison — current vs proposed quality preset:**
- ❌ Current (`generate_picture_images` + `generate_table_images`): **0 fixed + unbounded per-doc** cost. Memory spikes unpredictably with large/image-heavy PDFs. Images are never used since output is markdown. Pure waste.
- ✅ Proposed (`do_code_enrichment` + `do_formula_enrichment`): **~900 MB fixed** (CodeFormulaV2 model loaded once at startup) + negligible per-doc overhead. **Predictable, bounded memory.** Produces better markdown (code language tags, LaTeX formulas).

**Required change in `_download_models()`:**
```python
with_code_formula=True,  # currently False
```

**Note:** `do_picture_classification` does NOT require `generate_picture_images` — classification runs on internal page images during processing; `generate_picture_images` only controls whether extracted images are saved to the output document.

ramteid:
> @ramteid Would appreciate a review on this change when you get the chance.

Sure thing. I let Claude find some bugs and generate comments accordingly. But what I actually noticed myself is the issue with the Docling parameters. I use Docling at work and have some experience with these parameters. Not all of them add value in this use case and some are incompatible in general. Hope my comments help.

I assume OCR within images that are embedded in documents is currently overkill, right? Even on very good systems with GPU, Docling can get really, really slow with all parameters enabled. Especially VLM/granite can make conversion riddiculously slow.

ramteid (commented):
Second review pass — focusing on logic-level issues beyond the first round.

services/docling/app.py
```diff
@@ -1,309 +1,463 @@
...
+    try:
+        logger.info("Downloading models ...")
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(None, _download_models)
+
+        # Build one converter per preset. The ML models are shared via HF cache
+        # so subsequent converters are fast to construct.
+        for preset_name in QUALITY_PRESETS:
+            converter = await loop.run_in_executor(None, _build_converter, preset_name)
+            _converters[preset_name] = converter
```
ramteid: **Init fails atomically even when only one preset is broken**

If building the "quality" converter fails (missing model, download error, OOM), the whole `_init()` aborts and `_ready` stays `False`.  The "fast" and "balanced" converters that were already built and stored in `_converters` are never exposed — the service answers 503 to every request.

A more resilient approach: catch failures per-preset inside the loop, log a warning, and set `_ready = True` as long as at least one converter succeeded.  Requests for a failed preset can return 400 with a clear message instead of the entire service going dark.

services/docling/app.py
```diff
@@ -1,309 +1,463 @@
...
+        finally:
+            _cleanup_result(result)
+
+    # Large PDF: chunked conversion
+    markdown_parts: list[str] = []
+    for start in range(1, page_count + 1, _PAGES_PER_CHUNK):
+        end = min(start + _PAGES_PER_CHUNK - 1, page_count)
+        logger.debug("Processing pages %d–%d of %d", start, end, page_count)
+        stream = DocumentStream(name=filename, stream=io.BytesIO(data))
+        result = converter.convert(stream, page_range=(start, end))
```
ramteid: **Cross-page elements are corrupted at chunk boundaries**

When a table, figure, or numbered list straddles the boundary between two chunks (e.g. starts on page 5 and continues on page 6 with `_PAGES_PER_CHUNK=5`), each chunk is fed to Docling independently.  The first chunk emits a truncated table; the second starts with an orphan continuation.  `"\n\n".join()` pastes them together with no awareness of this.

For typical business documents with multi-page tables this will produce visibly broken markdown.  Two possible mitigations:
1. Add a 1-page overlap between chunks and de-duplicate the markdown.
2. Pre-split the PDF into physical per-chunk PDFs with `pypdfium2` (which you already import), so Docling only ever sees the pages it needs — this also avoids the full-file reparse problem below.

services/docling/app.py
```diff
@@ -1,309 +1,463 @@
...
+            return result.document.export_to_markdown()
+        finally:
+            _cleanup_result(result)
+
+    # Large PDF: chunked conversion
+    markdown_parts: list[str] = []
+    for start in range(1, page_count + 1, _PAGES_PER_CHUNK):
+        end = min(start + _PAGES_PER_CHUNK - 1, page_count)
+        logger.debug("Processing pages %d–%d of %d", start, end, page_count)
+        stream = DocumentStream(name=filename, stream=io.BytesIO(data))
```
ramteid: **Every chunk re-reads and re-parses the entire PDF**

Each iteration creates `DocumentStream(name=..., stream=io.BytesIO(data))` from the full file bytes and hands it to Docling.  Docling/pypdfium2 then parses the complete PDF just to extract 5 pages for the ML pipeline.  For a 200 MB / 100-page document this means 20 full parses.

Since you already import pypdfium2 for the page count, you could split the PDF once up front into small per-chunk PDFs and pass only those chunks to `converter.convert()`.  That would cut both parse time and peak memory significantly.

services/docling/app.py
```diff
@@ -1,309 +1,463 @@
...
+                and j.created_at < finished_cutoff
+            )
+        ]
+        for jid in stale:
+            del _jobs[jid]
+        if stale:
+            logger.info("Cleaned up %d stale job(s).", len(stale))
+
+
+async def _run_job(job: Job, data: bytes, filename: str) -> None:
```
ramteid, line 335: **No per-job timeout — one slow document can starve the whole service**

`_run_job` holds the semaphore for the entire duration of `_convert_document` with no upper bound.  A single pathological PDF (thousands of pages, complex layouts) can hold the permit for hours, during which every other submission ends up either queued or rejected with 429.

Consider wrapping the `run_in_executor` call in `asyncio.wait_for(…, timeout=MAX_JOB_SECONDS)` so a stuck conversion is cancelled after a configurable limit and the semaphore is released.

services/docling/app.py
```diff
@@ -1,309 +1,463 @@
...
+
+@app.get("/jobs/{job_id}")
+async def get_job(job_id: str):
+    """Poll a conversion job. Returns status and, when complete, the Markdown result."""
+    job = _jobs.get(job_id)
+    if job is None:
+        raise HTTPException(status_code=404, detail=f"Unknown job ID: {job_id!r}")
+    if job.status == JobStatus.COMPLETED:
+        logger.info("Job %s: result retrieved.", job_id)
+        del _jobs[job_id]
```
ramteid: **Job result deleted before the client receives the response**

`del _jobs[job_id]` runs inside the request handler, *before* FastAPI serialises and transmits the response.  If the connection drops after the delete but before the client reads the body, the result is permanently lost — the next poll returns 404.

Safer option: don't delete on read.  Let `_cleanup_loop` handle all deletions.  A completed job stays in `_jobs` for at most 10 minutes, which is acceptable given `_MAX_PENDING` is small.

services/connector-manager/src/handlers.rs
```diff
@@ -829,10 +834,31 @@ async fn is_docling_enabled(redis_client: &redis::Client) -> bool {
...
+            return "balanced".to_string();
+        }
+    };
+
+    let value: Option<String> = conn
+        .hget(REDIS_SYSTEM_SETTINGS_KEY, "docling_quality_preset")
+        .await
+        .ok();
+    value.unwrap_or_else(|| "balanced".to_string())
+}
```
ramteid, line 889: **Preset read from Redis is never validated**

`get_docling_quality_preset()` returns whatever string is stored in Redis.  If the value is corrupted, stale, or set to something like `"ultra"`, it passes straight through to Docling, which returns 400 → the connector-manager silently falls back to built-in extraction via the `Err(e) => None` arm in `do_extract_text`.  The admin has no indication their quality setting is broken.

Validate against a known set here and fall back to `"balanced"` with a warning log when the value is unrecognised.

prvnsmpth:
@ramteid Thanks for the review! I'm noticing some memory leak issues with docling, and apparently others have noticed this as well: https://github.com/docling-project/docling/issues/2209. Have you run into any such issues? 

I'm currently trying to index a few 1000 pdf/docx/pptx files and memory usage keeps growing (I've given docling 6g) and it finally OOMs and restarts.

I think you're right about the OCR, it's probably overkill for most use-cases. Let me try disabling it.

sdk/python/omni_connector/storage.py
prvnsmpth: I think this is fine as-is, it's very little duplicate logic

ramteid:
> @ramteid Thanks for the review! I'm noticing some memory leak issues with docling, and apparently others have noticed this as well: [docling-project/docling#2209](https://github.com/docling-project/docling/issues/2209). Have you run into any such issues?
> 
> I'm currently trying to index a few 1000 pdf/docx/pptx files and memory usage keeps growing (I've given docling 6g) and it finally OOMs and restarts.
> 
> I think you're right about the OCR, it's probably overkill for most use-cases. Let me try disabling it.

I wasn't aware of general OOM issues in Docling, that's bad. For the CPU mode, we have a memory limit anyway. But for GPU there is no VRAM limit. I am not quite sure how to circumvent this issue. A restart after each conversion probably could work, but we'd need to measure startup time. Another approach could be a watchdog service for RAM and VRAM usage that kills Docling before OOM happens. Not ideal though.

prvnsmpth:
@ramteid One solution is to run the docling conversion in a sub-process, that way we can kill the process each time we want to reclaim resources. But it's a little more complicated to implement.

PR #147, opened 2026-04-12
fix: web connector document_id mismatch and URL normalization

cuberone:
### Changes

**Bug fix: create/delete document_id mismatch**

`to_connector_event` used `base64(url)` as `document_id` while `publish_deletion_event` used `md5(url)`. Deletion events therefore never matched any indexed document. Both now use `base64(url)` consistently — the Redis URL set stores document IDs instead of MD5 hashes.

**URL normalization**

Added `WebPage::normalize_url` applied at `from_html` entry point:
- Strips `#fragment` (client-side only, not part of resource identity)
- Removes `utm_*` query params
- Scheme+host lowercasing handled automatically by the `url` crate

Prevents the same page from getting multiple Redis/DB rows when reached via URLs that differ only in tracking params or anchors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1328,6 +1328,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "conquer-once"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


cuberone:
Made by mistake

PR #149, opened 2026-04-12
feat: cross-source email deduplication for IMAP

ramteid:
## Problem

When two IMAP accounts subscribe to the same mailing lists, identical emails are ingested as separate documents because `external_id` (aka `document_id`) includes `source_id`. This degrades RAG quality by diluting the result set with near-duplicate hits at different scores.

## Solution

Remove `source_id` from IMAP thread document IDs so that the same mailing-list thread ingested from different accounts produces the same `external_id`. This enables deduplication across the entire RAG pipeline without any schema changes.

### Connector (`connectors/imap/src/models.rs`)
- `make_thread_document_id(folder, thread_id)` no longer includes `source_id`
- Produces e.g. `imap-thread:INBOX:<thread@example.com>`
- Folder stays in the ID to keep Gmail's `INBOX` vs `[Gmail]/All Mail` distinct and to avoid `UNIQUE(source_id, external_id)` violations within the same source

### Search-time deduplication (`services/searcher/src/search.rs`)
- Over-fetches with `dedup_limit()` (3× requested limit, capped at 300) on fulltext, semantic, and hybrid search paths
- `deduplicate_cross_source()` groups results by `external_id` for `imap-thread:` prefix only — other content types pass through unchanged
- Keeps the highest-scoring result per group (tiebreaker: `updated_at` for most recent sync)
- Collapsed duplicates are recorded in `also_in: [{source_id, document_id, score}]` on the winning result for frontend attribution
- Permissions are **not** mutated — the SQL layer already filters by the requesting user's access
- **`total_count` is adjusted** by subtracting the number of collapsed duplicates so that pagination reflects the actual number of unique results and `has_more` is accurate

### Embedding cloning (`services/ai/`)
- `find_embedded_duplicate(external_id, exclude_document_id)` locates a donor document with the same `external_id` that already has completed embeddings
- `clone_for_document(source_document_id, target_document_id)` copies embedding rows with fresh ULIDs instead of re-embedding identical content
- `batch_processor` checks for a donor before fetching content, saving compute
- **Cloning is restricted to `imap-thread:` prefix only** — other source types that may share `external_id` values in the future are not affected

### Data migration (`services/migrations/083_rewrite_imap_thread_external_ids.sql`)
- Rewrites existing old-format `external_id` values (`imap-thread:{source_id}:{folder}:{thread_id}`) to new format (`imap-thread:{folder}:{thread_id}`)
- Matches rows by verifying the embedded segment equals the row's own `source_id` — avoids false positives on new-format rows
- Handles collision gracefully: if the updated connector already inserted a new-format row, the old-format duplicate is deleted (CASCADE removes its embeddings and queue entries)

## Files changed

| Area | Files |
|------|-------|
| IMAP connector | `connectors/imap/src/models.rs`, `connectors/imap/src/sync.rs`, `connectors/imap/tests/integration_test.rs` |
| Searcher | `services/searcher/src/search.rs`, `services/searcher/src/models.rs` |
| AI service | `services/ai/db/documents.py`, `services/ai/db/embeddings.py`, `services/ai/embeddings/batch_processor.py` |
| Migration | `services/migrations/082_rewrite_imap_thread_external_ids.sql` |

## What's NOT changed
- No schema changes — leverages existing `idx_documents_external_id` index and `UNIQUE(source_id, external_id)` constraint
- No changes to shared models, SDK, or indexer
- Non-IMAP connectors are completely unaffected

## Caveats
- Different folder names across accounts (e.g. `INBOX` vs `Archive`) will not collapse — acceptable since mailing-list dedup targets the same folder
- Embeddings and queue entries for collision victims are CASCADE-deleted by the migration

## Testing
- IMAP unit tests: 8/8 pass (including cross-source ID equality test)
- IMAP integration tests: 6/6 pass
- Searcher compiles clean with no new warnings
- Compilation clean across all packages after rebase on upstream/master


services/migrations/082_add_content_fingerprint_to_documents.sql
```diff
@@ -0,0 +1,9 @@
+-- Add content_fingerprint column for cross-source deduplication.
+-- Connectors compute a fingerprint from source-independent identifiers
+-- (e.g. RFC 2822 Message-ID for email threads) so the indexer and AI
+-- service can detect semantically identical documents across sources
+-- without relying on content hashing (which fails when bodies differ
+-- due to mailing-list footers, DMARC From-rewriting, per-account flags, etc.).
+
+ALTER TABLE documents ADD COLUMN IF NOT EXISTS content_fingerprint VARCHAR(128);
```
prvnsmpth: I think we should use `external_id` for this purpose.

Currently imap constructs `external_id` like this:

```
pub fn make_thread_document_id(source_id: &str, folder: &str, thread_id: &str) -> String {
    format!(
        "imap-thread:{}:{}:{}",
        source_id,
        urlenc(folder),
        encode(thread_id)
    )
}
```

If we simply remove `source_id` from this construction, we will get cross-source unique IDs for each thread. Then, at query time, we can collapse results with the same external_id.

The problem with introducing a new field for this in documents is that it doesn't quite generalize well to other connectors. It's not clear what content_fingerprint should be used when indexing other types of content.

prvnsmpth (commented):
I think the `content_fingerprint` idea doesn't generalize well to all connectors. Using `external_id` for this seems to be a much simpler solution.

I've started a discussion thread on this: https://github.com/getomnico/omni/discussions/150 with a possible solution. Please let me know if this makes sense.

services/migrations/082_add_content_fingerprint_to_documents.sql
ramteid: Agreed, I made / reverted some changes. See my comment at #150

services/migrations/083_rewrite_imap_thread_external_ids.sql
```diff
@@ -0,0 +1,45 @@
...
+--
+-- The WHERE clause limits the update to rows whose external_id actually
+-- contains their own source_id (guaranteeing old-format; avoids false matches
+-- on new-format rows whose folder happens to be 26 alphanumeric chars).
+-- A CTE is used so that ON CONFLICT (source_id, external_id) collisions
+-- (possible if new-format rows were already inserted by the updated connector)
+-- are handled gracefully: conflicting rows are deleted in favour of the
+-- already-existing new-format row.
+
+WITH to_rewrite AS (
```
prvnsmpth, line 18: Do we really need to check-in this migration? I suppose the IMAP can be synced from scratch to resolve this?

ramteid: I guess - but how would you initialize re-sync for existing deployments?

prvnsmpth: I didn't think that there would be many other deployments :)  

But you're right, imap was released in 0.1.2, so any deployments at that version will need to be handled. Let's keep the migration to handle those. Thanks!

prvnsmpth:
Thanks for addressing the comments! Looks great, merged.

PR #151, opened 2026-04-13
fix(google): correct OAuth permission handling for personal Drive accounts

cuberone:
## Problem

When a user connects their Google Workspace account via OAuth, synced documents were indexed with **empty or incorrect permissions**, making them unsearchable by the authenticated user.

This manifested as three distinct bugs discovered through end-to-end testing of the multi-user OAuth flow.

---

### 1. OAuth credentials stored via wrong endpoint (callback → 403)

**File:** `web/src/routes/api/connectors/google/oauth/callback/+server.ts`

The OAuth callback was sending credentials to `${INDEXER_URL}/service-credentials`. Two problems:
- That endpoint does not exist on the indexer — returns **404**
- The correct endpoint (`/api/service-credentials`) requires **admin role** — returns **403** for regular users completing the OAuth flow

**Fix:** Relaxed the auth check on `POST /api/service-credentials` to allow the source owner in addition to admins. The callback now calls the API correctly using camelCase field names. Also fixed a related issue: the original call was sending snake_case (`source_id`, `auth_type`, `principal_email`) while the API expects camelCase.

```typescript
// Before: broken API call to wrong service with wrong field names
const indexerResponse = await svelteFetch(`${env.INDEXER_URL}/service-credentials`, {
    method: 'POST',
    body: JSON.stringify({ source_id: ..., auth_type: ..., principal_email: ... }),
})

// After: correct API call, owner allowed by service-credentials endpoint
const credResponse = await svelteFetch('/api/service-credentials', {
    method: 'POST',
    headers: { 'Content-Type': 'application/json' },
    body: JSON.stringify({ sourceId: ..., authType: ..., principalEmail: ... }),
})
```

**File:** `web/src/routes/api/service-credentials/+server.ts`

```typescript
// Allow source owner in addition to admins (e.g. OAuth callback for non-admin users)
const isOwner = source.createdBy === locals.user.id
if (locals.user.role !== 'admin' && !isOwner) {
    throw error(403, 'Forbidden')
}
```

---

### 2. Drive API fields request omitted `owners`, causing empty permissions

**File:** `connectors/google/src/drive.rs`

The `files.list` and `changes.list` API calls did not include `owners` in the `fields` parameter. For Google Workspace accounts, the file owner's access is **implicit** — the `permissions` array does not include an entry for the owner. Without `owners`, there was no way to identify who owned the file.

**Fix:** Added `owners(emailAddress)` to both field requests (initial sync and incremental/changes sync).

```rust
// Before
"fields", "nextPageToken,files(...,permissions(id,type,emailAddress,role))"

// After
"fields", "nextPageToken,files(...,permissions(id,type,emailAddress,role),owners(emailAddress))"
```

---

### 3. `Owner` struct `id` field non-optional caused deserialization failure

**File:** `connectors/google/src/models.rs`

After adding `owners(emailAddress)` to the fields request, the sync began failing with:

```
missing field `id` at line 7 column 9
```

The `Owner` struct had `id: String` (required), but when requesting only `owners(emailAddress)` the API does not return `id`. Deserialization failed on every file, the sync treated all files as deleted, and documents were removed from the index.

**Fix:** Changed `id` to `Option<String>`.

```rust
// Before
pub struct Owner { pub id: String,

// After
pub struct Owner { pub id: Option<String>,
```

---

### 4. Owner email not included in document permissions

**File:** `connectors/google/src/models.rs`

Even after getting the `owners` array from the API, the permission-building logic in `to_connector_event` never used it. For a Google Workspace Drive file with no explicit shares, the `permissions` array is completely empty — owner access is implicit. Result: `{"users": [], "groups": [], "public": false}` — the owner themselves could not find their own files.

**Fix:** After processing the `permissions` array, iterate `owners` and add each owner's email to `users` if not already present.

---

### 5. Viewer's email absent from permissions (non-owner OAuth sync)

**File:** `connectors/google/src/models.rs`, `connectors/google/src/sync.rs`

When an OAuth user syncs a file they have **viewer access** to (owned by someone else), the Google Drive API returns an **empty `permissions` array** for non-owners — only the file owner can enumerate the full permission list. The result was:

```json
{"users": ["owner@example.com"], "groups": [], "public": false}
```

The viewer (`viewer@example.com`) could not find the file in search even though they had explicit access.

**Logic:** If a file appears in a user's Drive listing, they have access to it by definition. The OAuth credentials used to fetch the listing are scoped to that user.

**Fix:** Added `oauth_user_email: Option<&str>` parameter to `to_connector_event`. When provided, the email is added to `users` (deduplicated). At the call site in `sync.rs`, pass `Some(&user_file.user_email)`.

```rust
// Viewers can't see the permissions array, but presence in Drive listing implies access
if let Some(email) = oauth_user_email {
    let email = email.to_string();
    if !users.contains(&email) {
        users.push(email);
    }
}
```

---

## Testing

Verified end-to-end with two Google Workspace accounts:

1. **Owner** (`owner@example.com`) owns a `.docx` file, shares it with **Viewer** (`viewer@example.com`)
2. Viewer connects Google Drive via OAuth → source created, credentials stored correctly
3. Admin triggers sync for viewer's source → document indexed with `{"users": ["owner@example.com", "viewer@example.com"]}`
4. Viewer searches → document appears ✓
5. Owner removes share → re-sync → viewer's email removed from permissions → document no longer appears in viewer's search ✓
6. Owner restores share → re-sync → document reappears ✓

### DIFF
--- a/connectors/google/src/drive.rs
+++ b/connectors/google/src/drive.rs
@@ -103,7 +103,7 @@ impl DriveClient {
 
             let mut params = vec![
                 ("pageSize", "100"),
-                ("fields", "nextPageToken,files(id,name,mimeType,webViewLink,createdTime,modifiedTime,size,parents,shared,permissions(id,type,emailAddress,role))"),
+                ("fields", "nextPageToken,files(id,name,mimeType,webViewLink,createdTime,modifiedTime,size,parents,shared,permissions(id,type,emailAddress,role),owners(emailAddress))"),
                 ("q", query.as_str()),
                 ("orderBy", "modifiedTime desc"),
                 ("includeItemsFromAllDrives", "true"),
@@ -787,7 +787,7 @@ impl DriveClient {
             ("includeItemsFromAllDrives", "true"),
             ("supportsAllDrives", "true"),
             ("includeRemoved", "true"),
-            ("fields", "nextPageToken,changes(changeType,removed,file(id,name,mimeType,webViewLink,createdTime,modifiedTime,size,parents,shared,permissions(id,type,emailAddress,role)),fileId,time)"),
+            ("fields", "nextPageToken,changes(changeType,removed,file(id,name,mimeType,webViewLink,createdTime,modifiedTime,size,parents,shared,permissions(id,type,emailAddress,role),owners(emailAddress)),fileId,time)"),
         ];
 
         let response = self
--- a/connectors/google/src/models.rs
+++ b/connectors/google/src/models.rs
@@ -45,7 +45,7 @@ pub struct GoogleDriveFile {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Owner {
-    pub id: String,
+    pub id: Option<String>,
     #[serde(rename = "emailAddress")]
     pub email_address: Option<String>,
     #[serde(rename = "displayName")]
@@ -128,6 +128,7 @@ impl GoogleDriveFile {
         source_id: &str,
         content_id: &str,
         path: Option<String>,
+        oauth_user_email: Option<&str>,
     ) -> ConnectorEvent {
         let mut is_public = false;
         let mut users = Vec::new();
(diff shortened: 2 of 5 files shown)


prvnsmpth:
Merged, thanks for the fix!

PR #166, opened 2026-04-18
fix: content-address blob storage + non-blocking GC

xingfanxia:
## Summary

Two related bugs in the content-blob storage path cause unbounded table growth and can permanently stall the indexer. Observed in production (see below).

### 1. `store_content_with_type` never deduplicates

`ContentStorage::store_content_with_type`, `PostgresStorage::store_content_with_type`, and `S3Storage::store_content_with_type` all generate a fresh ULID and `INSERT` a new `content_blobs` row on every call — the `sha256_hash` is computed but only stored, never consulted. Connectors that re-store identical content (e.g. a full-sync connector running hourly) grow `content_blobs` linearly with sync frequency rather than with unique content.

On our deployment:

| metric | value |
| --- | --- |
| content_blobs rows | 13.4M |
| distinct sha256 hashes | 188K |
| **duplicate factor** | **71×** |
| top-duplicated single blob | 60,589 copies |

The existing `find_by_hash` helper (`postgres.rs:165`) is dead code — it's defined but no writer calls it. The existing `test_content_deduplication` locks the buggy behavior in with `assert_ne!(content_id1, content_id2)`.

### 2. `ContentBlobGC::run` blocks the queue processor's main `select!`

`services/indexer/src/queue_processor.rs` runs `tokio::select!` inside its main loop with one branch for `gc_interval.tick()`. That branch calls `gc.run().await` inline, and `gc.run()` internally loops `fetch_expired_orphans → delete` until the backlog is empty. In production an 8M orphan backlog held that branch for 8 days, during which every other branch — notification listener, `check_interval` batching, heartbeat, stale-recovery — stopped firing. The indexer looked healthy (container up, /health returning 200) but processed zero events.

## The fix

**`shared/src/storage/{postgres,s3}.rs` + `shared/src/content_storage.rs`** — look up by hash first, only `INSERT` when the content is new. The concurrent `SELECT`+`INSERT` race can still produce a small bounded number of duplicates per hash, but that's ~O(writer concurrency) instead of O(sync_count), and they get cleaned up by the orphan GC. No schema migration required.

**`services/indexer/src/queue_processor.rs`** — spawn GC on its own tokio task via `Semaphore(1).try_acquire_owned()`, so overlapping ticks are skipped rather than queued and the main `select!` is never held.

**`shared/src/storage/gc.rs`** — cap `delete_expired_orphans` at 50K blobs per invocation so even a very large backlog is drained across successive scheduled runs instead of in one marathon sweep.

**Tests** — `test_content_deduplication` now asserts `assert_eq!(id1, id2)`, matching what content-addressed storage is supposed to do.

## Evidence from prod deployment

After deploying these changes to our instance:
- **19,000 events processed over 3 minutes**, **0 new `content_blobs` rows** (100% dedup hit rate)
- Indexer main loop unblocked; queue draining at ~875 docs/sec
- No GC-related slow-statement warnings in logs
- Prior write-rate of ~800 new blobs/min → 0/min

## Test plan

- [x] `cargo check -p shared -p omni-indexer` passes (rustc 1.91.1)
- [x] `test_content_deduplication` updated to assert dedup behavior
- [x] Deployed to production, verified 0 new blobs over 3-min sampling window while 19K events processed
- [x] No regressions in batch processing throughput
- [ ] Reviewer may want to add a test that stops GC from running past a per-call cap under simulated backlog

## Follow-ups not in this PR

- Existing `content_blobs` tables with heavy duplication still need a one-time dedup pass + a unique index on `sha256_hash` to make the guarantee strict (the find-by-hash path still races). I'd do that as a separate migration once the fix is in.
- Separately, the HubSpot connector appears to re-emit the same object text many times per hour, which is wasteful even when dedup short-circuits the write. Worth investigating but not in scope here.

### DIFF
--- a/shared/src/storage/postgres.rs
+++ b/shared/src/storage/postgres.rs
@@ -32,14 +32,28 @@ impl ObjectStorage for PostgresStorage {
         content_type: Option<&str>,
         _prefix: Option<&str>,
     ) -> Result<String, StorageError> {
-        let content_id = generate_ulid();
         let size_bytes = content.len() as i64;
 
-        // Generate SHA256 hash for potential deduplication
         let mut hasher = Sha256::new();
         hasher.update(content);
         let hash = format!("{:x}", hasher.finalize());
 
+        // Content-address: reuse existing blob when hash matches. Under concurrent
+        // writes the SELECT+INSERT race may produce a small bounded number of
+        // duplicates per hash; those are cleaned up by the orphan GC.
+        let existing: Option<String> = sqlx::query_scalar(
+            "SELECT id FROM content_blobs WHERE sha256_hash = $1 LIMIT 1",
+        )
+        .bind(&hash)
+        .fetch_optional(&self.pool)
+        .await
+        .map_err(|e| StorageError::Backend(format!("Failed to lookup content by hash: {}", e)))?;
+
+        if let Some(id) = existing {
+            return Ok(id);
+        }
+
+        let content_id = generate_ulid();
         sqlx::query(
             r#"
             INSERT INTO content_blobs (id, content, content_type, size_bytes, sha256_hash, storage_backend)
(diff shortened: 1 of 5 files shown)


shared/src/storage/postgres.rs
```diff
@@ -32,14 +32,28 @@ impl ObjectStorage for PostgresStorage {
...
     ) -> Result<String, StorageError> {
-        let content_id = generate_ulid();
         let size_bytes = content.len() as i64;
 
-        // Generate SHA256 hash for potential deduplication
         let mut hasher = Sha256::new();
         hasher.update(content);
         let hash = format!("{:x}", hasher.finalize());
 
+        // Content-address: reuse existing blob when hash matches. Under concurrent
```
prvnsmpth, line 41: This is great! Will help with duplicate content a lot in certain connectors.

PR #168, opened 2026-04-18
Memory Integration

cuberone:
# Memory

The AI service embeds [mem0ai](https://github.com/mem0ai/mem0) in-process to provide long-term memory for chat and background agents.

## Why this exists

Memory allows chat and agents to persist context across sessions. It is treated as **non-critical infrastructure**: if memory fails to initialize or a call fails, the AI service degrades gracefully to no-memory behaviour. Features like chat and agents must never break entirely due to a memory failure.

## Components

The memory implementation lives in `services/ai/memory/`:

```
services/ai/memory/
├── bootstrap.py      # Builds mem0 configuration from AI service state (LLM, embedder)
├── mode.py           # Resolves the effective memory mode (off / chat / full)
├── role_bootstrap.py # Bootstraps the restricted mem0ai Postgres role
└── service.py        # Direct async wrapper around mem0.Memory, mimicking the old HTTP client
```

### `service.py`

This provides the core `MemoryService` interface, replacing the old external HTTP memory service. It exposes asynchronous methods that internally dispatch to mem0 via `fastapi.concurrency.run_in_threadpool` because mem0 is synchronous. 

The API surface matches what the `routers/memory.py` proxy and other internal callers expect:
- `add()`: Add messages to memory
- `search()`: Vector-search memories matching a query
- `list()`: List every memory stored for a `user_id`
- `delete()`: Delete a single memory by ID
- `delete_all()`: Delete every memory for a `user_id`

`user_id` is an opaque scoping key used for namespacing.

### Memory Namespaces & Trust Levels

The AI service uses different key shapes (`user_id` equivalent) for different subjects:

| Namespace | Rendered as | Why |
| --------- | ----------- | --- |
| `<user_id>` (chat) | `<untrusted-memory>` fence with safety contract | Content is extracted from connector data (Slack, Gmail, etc.) — treated as attacker-controlled |
| `user:<uid>:agent:<id>` (personal agent) | Plain `## What I remember` bullet list | Derived from the agent's own instructions and run summaries — not user-controlled data |
| `org_agent:<agent_id>` (org agent) | Plain `## Agent memory (from prior runs)` bullet list | Same rationale as personal agent |

The fence in chat memory instructs the model to treat the content as observations only, not instructions. Agent memory has no fence because it originates from admin-controlled instructions and the agent's prior run summaries.

## How it's wired into the rest of Omni

```
browser
  │
  ▼
web (SvelteKit)
  │  x-user-id: <session user>
  ▼
ai service (:3003)
  ├── routers/chat.py    ── search + add for chat turns
  ├── routers/memory.py  ── session-scoped proxy (list/delete)
  ├── agents/executor.py ── search + add for agent runs
  └── memory/
        ├── service.py   ── in-process mem0 wrapper
        ├── bootstrap.py ── mem0 configuration logic
        └── mode.py      ── resolve effective mode
  │
  ▼
Postgres + pgvector (shared main omni DB, accessed via mem0ai role)
```

### AI service proxy (`routers/memory.py`)

Browsers and the web backend talk to the AI service's `/memories` router, which enforces ownership and authorization before forwarding calls to the in-process `MemoryService`:

| Method | Path                                | Auth  | Purpose                                                                      |
| ------ | ----------------------------------- | ----- | ---------------------------------------------------------------------------- |
| GET    | `/memories`                       | user  | List caller's memories (scoped to `x-user-id`)                             |
| DELETE | `/memories`                       | user  | Delete all of caller's memories                                              |
| DELETE | `/memories/org-agent/{agent_id}`  | admin | Purge `org_agent:<id>` namespace — called on org agent delete             |
| DELETE | `/memories/user-agent/{agent_id}` | admin | Purge `user:<uid>:agent:<id>` namespace — called on personal agent delete |
| POST   | `/memories/agent/{agent_id}/seed` | admin | Seed agent memory from instructions — called on agent create/update         |
| DELETE | `/memories/{memory_id}`           | user  | Delete a single memory (ownership verified first)                            |

### Mode gate

Before ever calling the `MemoryService`, callers evaluate `memory.mode.resolve_memory_mode(user_mode, org_default)`:

- `org_default` is a **ceiling**, not a fallback. Users cannot exceed the organisation-wide setting the admin picked.
- `off` disables memory entirely for that request.
- `chat` enables memory for chat only.
- `full` additionally enables agent-run memory.

If the effective mode is `off`, the service is bypassed completely.

## Agent memory use-cases (UC-A and UC-B)

Background agents use memory in `full` mode only.

### UC-A — Personal agent (agent_type = "user")

A scheduled agent owned by a specific user. Its memory is **isolated** from the owner's chat memory.

| Step                            | What happens                                                                                                                                                                                            |
| ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Gate**                  | If `effective_mode != 'full'` — skip memory entirely                                                                                                                                                 |
| **Key**                   | `user:<agent.user_id>:agent:<agent.id>`                                                                                                                                                               |
| **Read** (before run)     | `search(query=agent.instructions, user_id=key, limit=5)` — retrieves the most relevant summaries from prior runs and injects them into the system prompt under `## Agent memory (from prior runs)` |
| **Write** (after summary) | `add([{role:"user", "content":"Agent task: {instructions}"}, {role:"assistant", "content":"Agent run summary: {summary}"}], user_id=key)` — fire-and-forget, non-blocking                                |

### UC-B — Org agent (agent_type = "org")

An organisation-wide agent that runs with access to all data, not scoped to any individual user.

| Step                            | What happens                                                            |
| ------------------------------- | ----------------------------------------------------------------------- |
| **Gate**                  | If `effective_mode != 'full'` — skip memory entirely                 |
| **Key**                   | `org_agent:<agent.id>`                                                |
| **Read** (before run)     | Same as UC-A: `search(query=agent.instructions, user_id=key, limit=5)` |
| **Write** (after summary) | Same turn structure as UC-A, fire-and-forget                            |

## Configuration

Runtime variables related to memory:

| Var | Purpose |
| --- | --- |
| `MEMORY_ENABLED` | Set to `true` (default) or `false` to toggle memory feature entirely |
| `MEM0AI_DATABASE_USER` | Postgres role name for mem0 (default `mem0ai`) |
| `MEM0AI_DATABASE_ROLE_PASSWORD` | Password for the mem0 role. Required if memory is enabled |
| `MEM0_HISTORY_DB_PATH` | SQLite history path (default `/tmp/mem0_history.db`) |

The chosen LLM and embedder configurations are inherited dynamically at startup from the admin settings available to the AI service. Supported providers:
- **LLM**: openai, openai_compatible, anthropic, gemini, bedrock, aws_bedrock
- **Embedder**: openai, local (TEI), jina, bedrock

## Database Isolation (`mem0ai` role)

Mem0 data is stored directly in the **main Omni Postgres database** (not a separate `omni_mem0` DB), but inside its own auto-created `mem0_memories_<fp>` tables in the `public` schema.

Isolation against the main Omni application data is enforced via a dedicated Postgres role: `mem0ai` (bootstrapped by `memory/role_bootstrap.py`).
- The `mem0ai` role has `CREATE` privileges on `public` (to bootstrap its mem0 tables).
- We explicitly `REVOKE ALL` on all standard Omni tables, sequences, and functions from the `mem0ai` role.
- We run `ALTER DEFAULT PRIVILEGES` to ensure future migrations don't accidentally leak table access to the `mem0ai` role.

This ensures mem0 functionality cannot access main application data, even in case of a prompt injection or bug.

## Switching embedding models

Each embedder configuration (provider + model + dimensions) produces a fingerprinted collection name (`mem0_memories_<sha256-12>`).

Switching embedders automatically creates a fresh collection — no data migration occurs and there are no dimension-mismatch crashes. The previous collection remains in Postgres but is no longer queried until the old embedder is restored.

`DELETE /memories?user_id=…` purges the user from **all** `mem0_memories*` tables, including stale ones from previous embedder configs.

## Operational notes

- **Storage**: vectors live in Postgres (pgvector) in the main Omni DB, but queried only via the `mem0ai` role.
- **History DB**: mem0 uses an SQLite event history. Path is configurable via `MEM0_HISTORY_DB_PATH` (defaults to `/tmp/mem0_history.db`). Compose mounts the `mem0_history` named volume at `/var/lib/mem0` so the file survives container restarts; vectors are in Postgres regardless.
- **Messages ring-buffer**: mem0 keeps the last 10 raw conversation messages in the SQLite history. Our `DELETE` call clears this buffer.
- **Failure mode**: In-process calls use exception swallowing. If the DB is down or mem0 is misconfigured, callers degrade to "no memories" smoothly without terminating the process.

## Development & Testing

Unit tests for memory components (`test_memory_role_bootstrap.py`, `test_memory_bootstrap.py`, `test_memory_service.py`) live in `services/ai/tests/unit/`. Since the new architecture is in-process, tests rely extensively on mocking (`unittest.mock`) to patch `mem0.Memory`, `psycopg`, and `sqlite3` driver behaviour, requiring no runtime LLM or real Postgres DB.


prvnsmpth:
@cuberone Thanks for working on this! I had started a simple memory implementation a while ago, but haven't published the PR yet.

That said, I'm wondering whether we need to pull in another dependency to implement memory. I looked through the mem0 code and found that they don't really have bm25 search over memories. They use vanilla full-text search in Postgres. mem0 is best suited for an application where you need to support multiple storage systems, multiple vector dbs, etc. but by design, Omni will only ever use postgres, pg_search and pgvector.

Because our stack is powered by paradedb, we can implement BM25 search over memories. We have a pipeline for computing and storing embeddings, which can be reused to embed memories and search over them.

So I think having a parallel implementation that does search, embedding lookup, scoring of memories, when we have our own search stack will, in the long run, not be a good place for us. I would very much prefer to have a single search stack for both data and memories. 

And the memory system itself isn't that complicated to implement - we could basically do it with a single table that stores memories keyed by user_id or agent_id, allowing read/write of arbitrary key-value pairs, create a bm25 and pgvector index.

WDYT?

cuberone:
Thanks @prvnsmpth. I share your concern about 2 parallel retrieval stacks. Mem0 claims bm25 in https://github.com/mem0ai/mem0/releases/tag/v2.0.0 but actually they do have only pgvector, you are right. Probably because short messages perform well with no bm25 support, dont know. Graph is off in this PR. So extra lib is still lightweight and fast.
My thoughts:
We have to move to our self made solution, your arguments are solid. The only thing is making deduplication, conflict resolution, etc. for enterprise ready solution might take a while. So I would propose to keep this solution as it works well (and polished), and develop replacement for mem0. I was intentionally focused on decoupling mem0 and it hides behind a small ~200-line sidecar with just five endpoints. So swap later is good choice as to me.
What do you think?

prvnsmpth:
@cuberone Ok that sounds good 👍 We can abstract out a common memory interface, and mem0 can be one implementation of it, which can be swapped out later if needed.

But can we use it as a library in `omni-ai` instead of adding another container? I think running it as a library would be much simpler. I would like to avoid adding another container to the stack. I think this is lightweight enough that it can run in omni-ai without issues.

prvnsmpth:
Read through the db initialization, and I think implementing memory within omni-ai is definitely feasible, and I think will be the cleaner option overall.

One that's a bit tricky about all of this is that mem0 probably manages it's own migrations on the new memories db.
Is it possible to use the same omni db for the mem0 tables as well? Or does it need to be separate? We can create a separate role for the memory manager and restrict read/write to only the memory tables.

cuberone:
Yes, good. My intention was to make minimal invasion to ai module and max scalability. But it seems memories not a high load feature.

prvnsmpth:
Right, it will just be a few reads/writes to the db during chat/agent execution. And maybe a background task to consolidate memories. So I think it will be better to implement it within omni-ai directly as a library.

cuberone:
@prvnsmpth done, pr description updated

services/migrations/084_add_memory_mode.sql
```diff
@@ -0,0 +1,25 @@
+-- 083_add_memory_mode.sql
+-- Add per-user memory mode preference and org-wide default.
+
+ALTER TABLE users
+    ADD COLUMN IF NOT EXISTS memory_mode TEXT
+    CHECK (memory_mode IN ('off', 'chat', 'full'));
```
prvnsmpth: I think it would be better to store user prefs in a separate table. Otherwise this table will become bloated with all the user preferences we will add in the future.

docker/docker-compose.dev.yml
```diff
@@ -85,6 +85,7 @@ services:
     environment:
       <<: *dev-database-url
       LOG_LEVEL: debug
+      MEM0AI_DATABASE_PASSWORD: mem0ai
```
prvnsmpth: Aren't we using the same db to store the mem0 tables? If so, then we shouldn't need a separate db password right?

services/ai/memory/bootstrap.py
```diff
@@ -0,0 +1,175 @@
...
+    if embedder_block["config"].get("embedding_dims"):
+        vector_store_config["embedding_model_dims"] = (
+            embedder_block["config"]["embedding_dims"]
+        )
+
+    return {
+        "vector_store": {"provider": "pgvector", "config": vector_store_config},
+        "llm":          _llm_block(llm_provider),
+        "embedder":     embedder_block,
+        "history_db_path": "/tmp/mem0_history.db",
```
prvnsmpth: Should this be in `/tmp`? Would be better to create a volume mount and use that, wouldn't it?

services/ai/memory/bootstrap.py
```diff
@@ -0,0 +1,175 @@
...
+        dims=embedder_block["config"].get("embedding_dims", 0),
+    )
+    fp = hashlib.sha256(fp_str.encode()).hexdigest()[:12]
+    collection_name = f"mem0_memories_{fp}"
+
+    vector_store_config: dict[str, Any] = {
+        "host":            database_host,
+        "port":            database_port,
+        "dbname":          database_name,
+        "user":            "mem0ai",
```
prvnsmpth: We shouldn't hardcode these, they should be read from env

services/ai/memory/service.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class MemoryService:
+    """Direct wrapper around mem0.Memory with the MemoryClient surface."""
+
+    def __init__(self, memory: Memory, db_config: dict[str, Any]):
+        self._mem = memory
+        self._db_config = db_config
+
+    async def add(self, messages: list[dict[str, Any]], user_id: str) -> None:
```
prvnsmpth: We should try to use concrete types as much as possible: e.g., instead of `list[dict[str, Any]]` we should use the canonical message type `MessageParam` (the message type in the anthropic API)

The current codebase already has several places where we use generic types, I'm trying to fix them in my PRs.

prvnsmpth (commented):
Sorry for the delay in reviewing, I have reviewed this partially, left a few comments. Will review the rest soon.

services/migrations/084_add_memory_mode.sql
cuberone: Agreed.

docker/docker-compose.dev.yml
cuberone: Same DB, separate role. The mem0ai Postgres role is the isolation boundary — it cannot read Omni's data even if mem0 itself is compromised. The password authenticates that login. Will rename the env var to MEM0AI_DATABASE_ROLE_PASSWORD

services/ai/memory/bootstrap.py
cuberone: It's a buffer for mem0 - vectors live in Postgres, so losing it on restart is harmless. But you're right that /tmp looks sketchy. I'll add a named volume in compose.

services/ai/memory/bootstrap.py
cuberone: Agree.

services/ai/memory/service.py
cuberone: Agreed

cuberone:
@prvnsmpth please see changes

prvnsmpth:
@cuberone I've pushed a commit to this branch with some changes (commit [e0f8231](https://github.com/getomnico/omni/pull/168/commits/e0f8231c42e1d8655ff9b45019eb61c01a97392e)), please let me know if it looks ok. Feel free to revert in case something is not right.

These are the changes I've made in my commit:
* a MemoryProvider interface that allows plugging in any memory implementation
* Put all mem0 implementation related stuff in a single python package to isolate it from the rest of the service. Also removed the named volume (I know I had suggested adding it, but on reviewing it I think it's ok to use tmp)
* Use a single table for storing both global and per-user configuration (instead of two separate user_preferences and configuration tables)
* and a few other changes to the API etc.

This is still a WIP, I have reviewed some of it, and it needs testing. But this is the general direction I think we should go in. Let me know what you think!

cuberone:
It's nice. I like it

PR #169, opened 2026-04-18
Simplify batching in indexer

prvnsmpth:
The batching logic in indexer is a bit complex: it attempts to balance live updates with full sync. For live updates, we might get sporadic update events in the queue, which we need to process asap. During full syncs, we'll get a large number of events in the queue in a short period of time, and we should batch them into large batch sizes to minimize writes.


xingfanxia (commented):
Thanks for the ping! Overall the direction looks great — a couple of observations from running #166/#170 in production.

## Wins I'd flag

**`dequeue_batch` is much healthier now.** The old `WITH target_sync_run AS (... GROUP BY sync_run_id ORDER BY COUNT(*) DESC LIMIT 1)` was producing 4–10s slow-statement warnings under our load (5M+ pending). A flat `ORDER BY id LIMIT N FOR UPDATE SKIP LOCKED` should be orders of magnitude faster — happy to confirm once we deploy.

**Removing the listener + accumulation state machine** simplifies the loop substantially. Agreed it was hard to reason about; the timer-driven model is much clearer.

## Concern: starvation bug from #170 comes back

This PR is branched off `1ac8f53` (before #170 merged), and the new `process_batch` keeps the unbounded outer `loop { dequeue; if empty break; process }`. Combined with `process_batch_safe().await` being invoked inline from the `poll_interval.tick()` branch, this reintroduces exactly the pathology #170 was fixing:

- `poll_interval` fires → branch body runs `process_batch_safe().await`
- `process_batch` loops forever because any active full-sync connector keeps the queue non-empty
- Branch body never returns → `select!` is stuck on that branch
- `gc_interval`, `heartbeat_interval`, `retry_interval`, `recovery_interval`, `cleanup_interval` — none of them fire

We saw this play out in production: indexer appeared healthy (containers up, `Batch processed N documents successfully` logging steadily) but zero `Queue stats` / `Starting content blob garbage collection` / `Retried N failed events` logs for 10+ hours. GC in particular is important because without it, orphans pile up indefinitely (which is how we got the 8M-blob backlog that started this whole debugging session).

## Suggested fix — tiny

Just cap the outer loop:

```rust
async fn process_batch(&self) -> Result<()> {
    const MAX_BATCHES_PER_CALL: usize = 3;
    let mut total_processed = 0;

    for _ in 0..MAX_BATCHES_PER_CALL {
        let events = self.event_queue.dequeue_batch(self.batch_size).await?;
        if events.is_empty() {
            if total_processed > 0 {
                info!("Finished all available events. Total: {}", total_processed);
            }
            return Ok(());
        }
        // ... existing processing ...
    }

    if total_processed > 0 {
        info!("Processed {} events this call (cap reached, continuing next tick)", total_processed);
    }
    Ok(())
}
```

With `poll_interval=30s` and `batch_size=2000`, 3 iterations gets you up to 6K events per call in ~6–15s depending on DB latency, and subsequent calls re-enter via the next poll tick. The main `select!` cycles every few seconds instead of once per unbounded drain, which is enough for GC + heartbeat to fire reliably.

Alternatively, you could lower `MAX_BATCHES_PER_CALL` to 1 — since `poll_interval` is now the sole driver, one batch per tick is the simplest possible model, at the cost of effective throughput being gated by the 30s cadence. Tradeoff worth weighing.

Happy to send this as a separate PR rebased on top of #169 once it lands, if that's easier than folding it in here.

---

Also: your PR body mentions the need to balance live updates (asap) with full sync (batch). With the listener removed, live updates now wait up to `poll_interval` (30s default). Did you consider keeping the PG `LISTEN` as a pure "wake up sooner" signal (no accumulation state, just a `tokio::select!` branch that kicks a tick)? Just checking whether the latency increase was intentional or a consequence of the simplification.

PR #214, opened 2026-04-28
Align mutable segments rows with indexer batch size

prvnsmpth:
With the current mutable segment size of 5000, we pay a heavy penalty on each query. Reducing this size to match the indexers default batch size so that we do not accumulate too many rows in the mutable segment. That way, the penalty stays in the 10s of ms range (instead of the current 7+ sec penalty).


prvnsmpth:
Tested with full sync of ~7k documents, no degradation in search perf. Latency <300ms. We'll try this config until we run into any issues and tune it further.

PR #216, opened 2026-04-28
Fix/sdk body and orphan dequeue

cuberone:
# Unblock document indexing: TS SDK body + orphan-dequeue lock

Two independent regressions that together stalled every TS-SDK connector's documents in the queue. Each is one line of real change. Both shipped in recent merges and went unnoticed because no test exercised the failing path end-to-end.

This PR has two atomic commits — one per bug, each with its own integration test — so they can be reviewed (or reverted) independently.

---

## 1. `fix(sdk-typescript): send { count: 1 } body on incrementScanned`

`SdkClient.incrementScanned` POSTs to `/sdk/sync/{id}/scanned` with no body:

```ts
async incrementScanned(syncRunId: string): Promise<void> {
  const response = await this.post(`/sdk/sync/${syncRunId}/scanned`);   // body undefined
  ...
}
```

`this.post(path, body?)` skips `JSON.stringify` when `body` is `undefined`, so the wire request has zero bytes.

But the server-side handler `sdk_increment_scanned` extracts `Json<SdkIncrementScannedRequest>`:

```rust
pub struct SdkIncrementScannedRequest {
    #[serde(default = "default_count")]
    pub count: i32,
}
fn default_count() -> i32 { 1 }
```

Axum's `Json` extractor calls `serde_json::from_slice(&body)`. An empty body fails to parse with:

```
EOF while parsing a value at line 1 column 0
```

…and the handler returns 400. The `#[serde(default)]` only fills a missing **field inside a parsed JSON object** — never a missing body. (`{}` would work; nothing does not.)

### Why it stayed hidden

The TS SDK's existing test (`tests/client.test.ts`) only asserted the path, not the body, so it passed against `msw`'s default lenient matcher. The handler was tightened by adding the `Json<T>` extractor in #160 (`acf01a0b "Refactor Google connector to use SDK client"`); the corresponding SDK update only landed in the Python and Rust SDKs.

Verified parity:
- Python SDK (`sdk/python/omni_connector/client.py`) sends `json={"count": 1}`.
- Rust SDK (`sdk/rust/src/client.rs`) sends a `count: i32` body.
- TS SDK was the lone holdout.

### Symptom in production

Connectors typically wrap the emit + counter pair in one `try/catch`:

```ts
try {
  await this.emitDocument(...)        // ① queue INSERT — succeeds
  await ctx.incrementScanned();       // ② 400 EOF — throws
} catch (e) {
  logger.warn(`Failed to emit ${id}`); // ③ misattributes ② as ①
}
```

So every document logs `Failed to emit <id>` even though the queue row was already committed. Operators see thousands of "failures" in the logs while the queue has the events sitting there waiting. We hit this the moment our Huly connector tried to sync a non-empty workspace.

### Fix

Send `{ count: 1 }`. One line. Tightened the test to assert the body, not just the path, so this can't silently regress again.

---

## 2. `fix(indexer): scope dequeue locks with FOR UPDATE OF q`

`shared::queue::dequeue_batch_orphans` issues:

```sql
WITH batch AS (
    SELECT q.id
    FROM connector_events_queue q
    LEFT JOIN sync_runs s ON q.sync_run_id = s.id
    WHERE q.status = 'pending'
    AND s.id IS NULL
    ORDER BY q.id
    LIMIT $1
    FOR UPDATE SKIP LOCKED
)
```

Postgres rejects this at parse time:

```
ERROR: FOR UPDATE cannot be applied to the nullable side of an outer join
```

`FOR UPDATE` (un-scoped) acquires locks on every table in the FROM list. The right side of a `LEFT JOIN` can be NULL — and our filter `s.id IS NULL` is **literally** the case where there's no `s` row to lock. From the Postgres docs:

> It is not possible for SELECT FOR UPDATE to be applied to the result of an outer join when no rows from the inner table match.

### Why it kills the entire indexer

`queue_processor::process_batch` runs orphan dequeue first, then per-sync-type dequeue:

```rust
// Process orphan events first
while batches_dequeued < MAX_BATCHES_PER_CALL && orphan_count > 0 {
    let events = self
        .event_queue
        .dequeue_batch_orphans(self.batch_size)
        .await?;                                        // ← `?` propagates Err
    ...
}

// THIS NEVER RUNS when there are orphans:
for (sync_type, reason) in ready {
    let events = self.event_queue.dequeue_batch_by_sync_type(...).await?;
    ...
}
```

The moment `orphan_count > 0`, `dequeue_batch_orphans` throws, `?` returns Err, `process_batch` exits, and the working `dequeue_batch_by_sync_type` is never even attempted.

So a single deleted source whose `sync_runs` rows cascade away **stops the indexer for every other source too**. We watched our production indexer go from `Completed: 1625` to `Completed: 1625` for hours while pending climbed past 270, with this error in the logs every minute:

```
ERROR Failed to process batch: error returned from database:
       FOR UPDATE cannot be applied to the nullable side of an outer join
```

### Why it stayed hidden

The existing tests in `shared/tests/event_queue_test.rs` only call `dequeue_batch` (the simple variant with no join). `dequeue_batch_orphans` and `dequeue_batch_by_sync_type` were added in #169 (`c589c34a "Simplify batching in indexer"`) without test coverage. Dev environments typically have no orphans (orphans appear in production after a source is deleted), so the bug was invisible locally.

### Fix

Scope the lock to just the queue rows:

```sql
LIMIT $1
FOR UPDATE OF q SKIP LOCKED   -- not bare FOR UPDATE
```

`OF q` tells Postgres to only lock rows from `q`. The `s` side is read-only metadata for the `IS NULL` filter — no lock needed, and locking it isn't possible on the nullable side anyway.

Applied the same `OF q` scoping to `dequeue_batch_by_sync_type` for symmetry. Its inner join lets bare `FOR UPDATE` parse today, but locking `sync_runs` rows is wasteful — concurrent `sync_run` status updates would block on locks they don't need.

Added integration tests for both dequeue variants. The orphan test is a strict regression — it fails at SQL parse time without the fix.

---

## Test plan

- [x] `cd sdk/typescript && npm test` — 48/48 pass; new body-content assertion in `incrementScanned` test.
- [x] `cargo test -p shared --test event_queue_test` — both new tests pass; the orphan test fails at SQL parse with bare `FOR UPDATE` and passes with `FOR UPDATE OF q`.

## Compatibility / migrations

None. Both changes are pure code; no schema, config, or wire-protocol changes. Old SDK clients (no body) talking to a new server still 400; new SDK clients (with body) talking to an old server (no `Json<T>` extractor) work fine — the server ignored the body before #160. So the rollout sequence is just "upgrade clients to match the server" and there's no flag-day.

## Related history

- Server-side regression: #160 (`acf01a0b`) added `Json<SdkIncrementScannedRequest>` to the handler; only Python and Rust SDKs were updated to match.
- Indexer regression: #169 (`c589c34a`) introduced `dequeue_batch_orphans` and `dequeue_batch_by_sync_type` without tests.

### DIFF
--- a/sdk/typescript/src/client.ts
+++ b/sdk/typescript/src/client.ts
@@ -184,7 +184,16 @@ export class SdkClient {
   }
 
   async incrementScanned(syncRunId: string): Promise<void> {
-    const response = await this.post(`/sdk/sync/${syncRunId}/scanned`);
+    // Send an explicit `{ count: 1 }` body. The connector-manager handler
+    // extracts `Json<SdkIncrementScannedRequest>`, which fails to parse an
+    // empty body even though the struct uses `#[serde(default)]` — the
+    // default applies to a missing field within a parsed JSON object, not
+    // to a missing body. Without this, every call returns 400 "EOF while
+    // parsing a value" and the catch block in connectors mis-logs the
+    // failure as if the preceding emit failed.
+    const response = await this.post(`/sdk/sync/${syncRunId}/scanned`, {
+      count: 1,
+    });
     if (!response.ok) {
       const text = await response.text();
       throw new SdkClientError(
--- a/sdk/typescript/src/client.ts
+++ b/sdk/typescript/src/client.ts
@@ -184,13 +184,6 @@ export class SdkClient {
   }
 
   async incrementScanned(syncRunId: string): Promise<void> {
-    // Send an explicit `{ count: 1 }` body. The connector-manager handler
-    // extracts `Json<SdkIncrementScannedRequest>`, which fails to parse an
-    // empty body even though the struct uses `#[serde(default)]` — the
-    // default applies to a missing field within a parsed JSON object, not
-    // to a missing body. Without this, every call returns 400 "EOF while
-    // parsing a value" and the catch block in connectors mis-logs the
-    // failure as if the preceding emit failed.
     const response = await this.post(`/sdk/sync/${syncRunId}/scanned`, {
       count: 1,
     });
--- a/sdk/typescript/tests/client.test.ts
+++ b/sdk/typescript/tests/client.test.ts
@@ -137,12 +137,14 @@ describe('SdkClient', () => {
   });
 
   describe('incrementScanned', () => {
-    it('calls correct endpoint', async () => {
+    it('calls correct endpoint with { count: 1 } body', async () => {
       let calledPath = '';
+      let capturedBody: unknown;
 
       server.use(
-        http.post(`${BASE_URL}/sdk/sync/:id/scanned`, ({ params }) => {
+        http.post(`${BASE_URL}/sdk/sync/:id/scanned`, async ({ params, request }) => {
           calledPath = `/sdk/sync/${params.id}/scanned`;
+          capturedBody = await request.json();
           return HttpResponse.json({ success: true });
         })
       );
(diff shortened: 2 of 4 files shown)


prvnsmpth:
Thanks for the fixes! Merged.

PR #217, opened 2026-04-28
# fix(sdk/ts): catch up to upstream — title shim, user filter, /updated, complete()

cuberone:
## Why

The TypeScript SDK has been silently out of sync with the connector-manager contract since the upstream merge `7c21fd10`. Four real bugs land in production for every TS-SDK-based connector:

1. **Every doc lands as `"Untitled"`.** The indexer reads only `metadata.title` and falls back to `"Untitled"` when missing ([`queue_processor.rs:1026`](services/indexer/src/queue_processor.rs)). The TS SDK's `serializeConnectorEvent` only writes `metadata` to the wire — `doc.title` is dropped. The Python SDK has had a partial shim since inception; the TS SDK never did.

2. **`user_filter_mode` / `user_whitelist` / `user_blacklist` / `source_type` are silently ignored.** `connector-manager` returns all four on `/sdk/source/:id/sync-config` ([`handlers.rs:1533-1541`](services/connector-manager/src/handlers.rs)), but the TS `fetchSourceConfig` typed only `config/credentials/connector_state`. They evaporated at the type boundary; `SyncContext` had no `shouldIndexUser`. Configuring a whitelist on a TS-connector source had **zero runtime effect**.

3. **`documents_updated` has been pinned at 0 since the upstream merge.** `7c21fd10` made `sdk_complete` a body-less status-flip handler and added `POST /sdk/sync/:id/updated` for accumulating the counter server-side. The Rust SDK migrated in the same merge ([`sdk/rust/src/context.rs:116`](sdk/rust/src/context.rs)). The TS SDK was not migrated — every TS-connector sync since `.11` reports `documents_updated = 0` regardless of how many docs actually moved.

4. **`ctx.complete(newState)` silently drops `newState`.** `sdk_complete` no longer parses the body. The TS client kept serializing `{new_state: ...}`; on a zero-emit sync the source's cursor never advances, leading to slower-and-slower delta scans over time.

## What changed

- **Title shim** — non-mutating spread in `emit`/`emitUpdated` copies `doc.title` into `metadata.title` when missing. Caller's `Document` is never modified. Broader than Python's conditional shim (Python skips bare-metadata docs — residual gap tracked separately).
- **User filter + `source_type`** — `fetchSourceConfig` now parses through `SdkSourceSyncDataSchema` (zod-validated, applies sane defaults). `SyncContext` exposes `shouldIndexUser(email)` (case-insensitive — mirrors Python; deliberately differs from Rust's case-sensitive equivalent) and a `sourceType` getter. `server.ts` plumbs all four fields into the constructor.
- **Explicit `incrementUpdated`** — added to both `SdkClient` and `SyncContext`. Mirrors the Rust SDK's API exactly: connector authors call it with whatever count semantics they want, errors propagate so the caller chooses the policy. The Linear connector now calls `ctx.incrementUpdated(1)` after each successful `emit`/`emitUpdated`.
- **`complete()` migration** — `client.complete` shrunk to a path-only POST (mirrors Rust SDK). `ctx.complete(newState?)` keeps its convenience signature but now routes `newState` through `updateConnectorState` (the path that actually persists) before the status flip — back-compat for connector callers, semantically correct.

## Coverage

- 20 new SDK tests in `sdk/typescript/tests/` (new `context.test.ts` file plus extensions to `client.test.ts` and `server.test.ts`). Every behavioural change is covered fail-first → pass.
- Full SDK suite goes from 57 → 77 passing tests; SDK + Linear builds clean.

## Verification after merge / deploy

- `SELECT title FROM documents WHERE source_id = '<linear-source-id>' LIMIT 5;` — non-`Untitled`.
- `SELECT documents_scanned, documents_processed, documents_updated FROM sync_runs WHERE source_id = '<linear-source-id>' ORDER BY started_at DESC LIMIT 1;` — `documents_updated` is non-zero.
- Set `user_filter_mode='whitelist'` on a test source; confirm `ctx.shouldIndexUser(email)` reflects it.
- `SELECT connector_state FROM sources WHERE id = '<linear-source-id>';` — `last_sync_at` advances after each sync (was frozen on zero-emit syncs).

## Follow-ups (out of scope)

- The Python SDK has the equivalent `documents_updated` gap (no `/updated` call) and the bare-metadata title-drop case — separate Python-SDK PR.

### DIFF
--- a/sdk/typescript/src/server.ts
+++ b/sdk/typescript/src/server.ts
@@ -13,6 +13,7 @@ import {
   createSyncResponseStarted,
   createSyncResponseError,
   ActionResponse,
+  type SdkSourceSyncData,
 } from './models.js';
 import { getLogger } from './logger.js';
 
@@ -119,7 +120,7 @@ export function createServer(connector: Connector): Express {
       return;
     }
 
-    let sourceData: import('./models.js').SdkSourceSyncData;
+    let sourceData: SdkSourceSyncData;
     try {
       sourceData = await getSdkClient().fetchSourceConfig(sourceId);
     } catch (error) {
--- a/sdk/typescript/src/server.ts
+++ b/sdk/typescript/src/server.ts
@@ -119,18 +119,9 @@ export function createServer(connector: Connector): Express {
       return;
     }
 
-    let sourceData: {
-      config: Record<string, unknown>;
-      credentials: Record<string, unknown>;
-      state: Record<string, unknown> | null;
-    };
+    let sourceData: import('./models.js').SdkSourceSyncData;
     try {
-      const data = await getSdkClient().fetchSourceConfig(sourceId);
-      sourceData = {
-        config: data.config,
-        credentials: data.credentials,
-        state: data.connector_state,
-      };
+      sourceData = await getSdkClient().fetchSourceConfig(sourceId);
     } catch (error) {
       const message = error instanceof Error ? error.message : String(error);
       if (message.includes('404')) {
@@ -148,10 +139,16 @@ export function createServer(connector: Connector): Express {
       getSdkClient(),
       syncRunId,
       sourceId,
-      sourceData.state ?? undefined,
+      sourceData.connector_state ?? undefined,
       syncMode,
       documentsScanned,
-      documentsUpdated
+      documentsUpdated,
+      {
+        sourceType: sourceData.source_type,
+        userFilterMode: sourceData.user_filter_mode,
+        userWhitelist: sourceData.user_whitelist,
+        userBlacklist: sourceData.user_blacklist,
+      }
     );
     activeSyncs.set(sourceId, ctx);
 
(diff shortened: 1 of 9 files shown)


sdk/typescript/src/server.ts
```diff
@@ -119,18 +119,9 @@ export function createServer(connector: Connector): Express {
       return;
     }
 
-    let sourceData: {
-      config: Record<string, unknown>;
-      credentials: Record<string, unknown>;
-      state: Record<string, unknown> | null;
-    };
+    let sourceData: import('./models.js').SdkSourceSyncData;
```
prvnsmpth: maybe we should import this at the top? 

`import type SdkSourceSyncData from './models'`

prvnsmpth (approved):
Looks good! Just one minor comment about the import

sdk/typescript/src/server.ts
cuberone: Agree

PR #262, opened 2026-06-03
fix(imap): reconnect IMAP session after connection loss mid-sync

ramteid:
## Problem

When the TCP connection to the IMAP server drops unexpectedly during a sync run (broken pipe, connection reset, unexpected EOF), all subsequent IMAP operations on the dead session fail. The connector catches each folder-level failure with a warning and continues, but since every remaining folder operation is performed on the same broken session, every subsequent folder also fails.

The end result is that only the folders processed before the connection was lost are synced. The connector still reports the run as successful and saves connector state. On the next incremental sync, the skipped folders may be re-attempted, but if the connection consistently drops at a particular point in the folder list the same folders can be missed repeatedly.

## Root cause

`execute_sync` holds a single `ImapSession` for the entire sync run and shares it across all folders. There was no mechanism to detect that the session had become unusable or to re-establish it.

## Fix

After a folder-level sync failure, inspect the error chain for known connection-loss indicators (broken pipe, connection reset, EOF, OS error 32/104). When such an indicator is found, attempt to re-establish the session before continuing to the next folder. If reconnection succeeds, the remaining folders are processed normally. If reconnection fails, the loop exits early rather than continuing to attempt operations on a broken session.

A small `is_connection_error` helper centralises the detection logic so it can be reused if needed in future.

### DIFF
--- a/connectors/imap/src/sync.rs
+++ b/connectors/imap/src/sync.rs
@@ -2,6 +2,7 @@ use anyhow::{Context, Result};
 use omni_connector_sdk::SyncContext;
 use omni_connector_sdk::{ConnectorEvent, SyncType};
 use std::collections::{HashMap, HashSet};
+use std::io;
 use tracing::{error, info, warn};
 
 use crate::client::ImapSession;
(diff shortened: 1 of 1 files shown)


connectors/imap/src/sync.rs
```diff
@@ -14,6 +14,20 @@ use crate::models::{
...
+/// re-established after a folder sync failure.
+fn is_connection_error(e: &anyhow::Error) -> bool {
+    let msg = format!("{:#}", e).to_lowercase();
+    msg.contains("broken pipe")
+        || msg.contains("connection reset")
+        || msg.contains("connection closed")
+        || msg.contains("unexpected eof")
+        || msg.contains("os error 32")
+        || msg.contains("os error 104")
+}
```
prvnsmpth, line 36: String matching is kinda brittle.

Can we do something like this:
```
            return matches!(
                io_err.kind(),
                io::ErrorKind::BrokenPipe
                    | io::ErrorKind::ConnectionReset
                    | io::ErrorKind::ConnectionAborted
                    | io::ErrorKind::NotConnected
                    | io::ErrorKind::UnexpectedEof
            );
```

prvnsmpth (commented):
Thanks for the PR! Just one minor comment

connectors/imap/src/sync.rs
ramteid: Good call. Replaced the string-matching with `io::ErrorKind`-based detection that walks the full `anyhow` error chain via `e.chain().any(|cause| cause.downcast_ref::<io::Error>().is_some_and(...))`. Also took the opportunity to add an early-exit in the per-UID fallback loop (from #263) so a dead connection propagates immediately rather than retrying all 50 UIDs before the reconnect triggers.

PR #266, opened 2026-06-03
fix(docker-compose): pass DOCLING_ENABLED to the web service

ramteid:
## Problem

The web frontend reads the `DOCLING_ENABLED` environment variable to decide whether to show the docling configuration section in the admin settings page. The variable was already documented in the `.env` template and forwarded to the connector-manager and docling services, but it was absent from the `web` service environment block in `docker-compose.yml`.

As a result, even when docling was running and fully functional, the settings UI offered no way to enable, disable, or configure it. Users had to manually set the Redis key to make any changes.

## Fix

Add `DOCLING_ENABLED: ${DOCLING_ENABLED:-false}` to the `web` service environment section, consistent with the pattern already used for `AGENTS_ENABLED` and `MEMORY_ENABLED`. The default of `false` ensures that deployments without docling are not affected.


ramteid:
Superseded by #267 which covers the same DOCLING_ENABLED docker-compose change along with two additional related fixes.

PR #268, opened 2026-06-03
fix(connector-manager): cancel running syncs when source is disabled

ramteid:
## Problem

When a user disables a connector source in the UI, any in-progress sync for that source continues running until it completes naturally. This wastes resources and can produce documents for a source the user has intentionally turned off.

## Solution

Add a new scheduler phase `cancel_syncs_for_inactive_sources` that runs on every tick. It queries for `running` sync_runs whose source has `is_active = false` or `is_deleted = true`, sends a best-effort cancel to the connector process, and marks the sync_run as `cancelled` in the database with the message `Source was disabled`.

This is consistent with the existing `detect_stale_syncs` phase pattern: connector cancel is fire-and-forget (warn on failure), but the DB state is always updated.

## Changes

- `services/connector-manager/src/sync_manager.rs`: new `cancel_syncs_for_inactive_sources()` method
- `services/connector-manager/src/scheduler.rs`: new tick phase calling the above

### DIFF
--- a/services/connector-manager/src/sync_manager.rs
+++ b/services/connector-manager/src/sync_manager.rs
@@ -505,6 +505,62 @@ impl SyncManager {
         }
     }
 
+    /// Cancel any running sync whose source has been disabled or deleted.
+    /// Runs on every scheduler tick so that disabling a source stops an
+    /// in-progress sync within one tick interval.
+    pub async fn cancel_syncs_for_inactive_sources(&self) -> Result<Vec<String>, SyncError> {
+        let running: Vec<(String, String, SourceType)> = sqlx::query_as(
+            r#"
+            SELECT sr.id, sr.source_id, s.source_type
+            FROM sync_runs sr
+            JOIN sources s ON sr.source_id = s.id
+            WHERE sr.status = 'running'
+            AND (s.is_active = false OR s.is_deleted = true)
+            "#,
+        )
+        .fetch_all(&self.pool)
+        .await
+        .map_err(|e| SyncError::DatabaseError(e.to_string()))?;
+
+        let mut cancelled = Vec::new();
+        for (sync_run_id, source_id, source_type) in running {
+            info!(
+                "Cancelling sync {} for inactive/deleted source {}",
+                sync_run_id, source_id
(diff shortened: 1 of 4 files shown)


services/connector-manager/src/sync_manager.rs
```diff
@@ -505,6 +505,62 @@ impl SyncManager {
...
+            WHERE sr.status = 'running'
+            AND (s.is_active = false OR s.is_deleted = true)
+            "#,
+        )
+        .fetch_all(&self.pool)
+        .await
+        .map_err(|e| SyncError::DatabaseError(e.to_string()))?;
+
+        let mut cancelled = Vec::new();
+        for (sync_run_id, source_id, source_type) in running {
```
prvnsmpth: I think we should keep all DB queries contained within repository struct implementations. This particular one should go in shared/src/db/repositories/sync_run.rs

And I think instead of having very specific methods like cancel_syncs_for_inactive_sources, we should have more generic and potentially reusable methods, e.g., fetch_runs(source_ids: &[String]), mark_cancelled(run_ids: &[String]). The current impl already has some very specific methods like get_running_for_source_in_types that we should ideally refactor into more generic methods.

In this particular case, we first determine disabled sources, then fetch the running sync runs for those sources, then call another method to mark them cancelled. This is 3 queries versus the single query above, but it's alright because these are tiny tables so perf is not really a concern.

ramteid: Done. Refactored in the follow-up commit (43a38130):

- Added `SourceRepository::find_inactive()` — queries sources where `is_active = false OR is_deleted = true`
- Added `SyncRunRepository::find_running_for_sources(source_ids: &[String])` — fetches running runs for a given set of source IDs
- Added `SyncRunRepository::mark_cancelled_many(ids: &[String], message: &str)` — batch UPDATE with `RETURNING id`, so only rows actually transitioned are counted
- `cancel_syncs_for_inactive_sources` now uses these three methods in sequence (find inactive sources → find their running runs → batch-cancel), with no inline SQL left in sync_manager.rs

PR #271, opened 2026-06-04
fix(imap): detect message deletions during full sync

ramteid:
## Problem

Full sync reset `connector_state` to `ImapConnectorState::default()`, which zeroed out `indexed_uids`.  With `indexed_uids = []`, the deletion diff (`indexed_uids − server_uids`) always produced an empty set — messages deleted from the server were silently orphaned in the Omni index and never cleaned up.

This affected anyone who triggered a full sync (e.g. via the UI "Force resync" button) after deleting messages.  The deleted messages would remain searchable in Omni indefinitely.

## Root cause

`connectors/imap/src/sync.rs`, `run_sync()`:

```rust
// BEFORE — full sync reset everything, preventing deletion detection
let mut connector_state = if ctx.sync_mode() == SyncType::Full {
    ImapConnectorState::default()   // ← indexed_uids = [] here
} else {
    state.unwrap_or_default()
};
```

## Fix

On a full sync, load the previous state and:

1. Capture `deleted_uids = indexed_uids − server_uids` (deletion diff, same as incremental)
2. **Clear `indexed_uids`** after capturing deletions — clean slate so the re-index pass has no duplicates
3. **Keep `messages` snapshots** — the deletion loop needs them to reconstruct thread document IDs for `DocumentDeleted` / `DocumentUpdated` events
4. **Clear `skipped_uids`** — full sync intent is to retry everything, including previously oversized messages
5. Set `new_uids = all server_uids` so every message body is re-downloaded

## Invariants preserved

| Property | Incremental | Full sync (before) | Full sync (after) |
|---|---|---|---|
| Deletions detected | ✅ | ❌ | ✅ |
| All messages re-fetched | ❌ | ✅ | ✅ |
| No duplicate indexed_uids | ✅ | ✅ | ✅ |
| Thread updates on partial delete | ✅ | ❌ | ✅ |

## Changes

- `connectors/imap/src/sync.rs`: fix `run_sync()` and `sync_folder()` as described above

### DIFF
--- a/connectors/imap/src/sync.rs
+++ b/connectors/imap/src/sync.rs
@@ -43,11 +43,22 @@ impl SyncManager {
         }
 
         let mut connector_state = if ctx.sync_mode() == SyncType::Full {
+            // On a full sync we want to re-fetch all messages from scratch, but
+            // deletion detection requires the previous state:
+            // - indexed_uids: so deleted_uids = old_indexed_uids − server_uids can be computed
+            // - messages: so the deletion loop can find thread snapshots to emit DocumentDeleted
+            // indexed_uids is cleared immediately after deleted_uids is computed in
+            // sync_folder, before new messages are pushed, to avoid duplicates.
+            // Clear skipped_uids so oversized messages are retried on a full sync.
             info!(
-                "Full sync requested, resetting connector state for source {}",
+                "Full sync requested, preserving state for deletion detection for source {}",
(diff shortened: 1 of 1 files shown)


ramteid:
Superseded by #273. Verified that #273 contains all code changes from #271 and adds one additional thread-cleanup safeguard described in #273.

PR #272, opened 2026-06-04
fix(ai): fix OpenAI provider for reasoning models (gpt-5.x)

ramteid:
## Problem

OpenAI reasoning models such as `gpt-5.2` reject sampling parameters like `temperature`, and small utility calls such as title generation can also exhaust `max_output_tokens` before visible text is produced. The result is failed title generation and opaque empty-response errors.

## Changes

- Detect OpenAI reasoning models (`gpt-5`, `o1`, `o3`, `o4`) and omit unsupported `temperature` / `top_p` parameters for them.
- Preserve sampling parameters for non-reasoning OpenAI models.
- Enforce a 1024-token floor for reasoning-model output budgets so small calls have room for visible output.
- Handle incomplete, failed, and cancelled Responses API statuses/events explicitly.
- Preserve user text when a message also contains tool results, and guard missing function-call item IDs.
- Add unit coverage for reasoning-model sampling omission and token floor behavior.

### DIFF
--- a/services/ai/providers/openai.py
+++ b/services/ai/providers/openai.py
@@ -32,6 +32,39 @@ from . import LLMProvider, LLMProviderStreamError, TokenUsage
 
 logger = logging.getLogger(__name__)
 
+MIN_REASONING_OUTPUT_TOKENS = 1024
+
+
+def _is_reasoning_model(model: str) -> bool:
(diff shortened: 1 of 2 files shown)


prvnsmpth:
Thanks for these fixes! Merged.

PR #274, opened 2026-06-04
fix(ai): parse connector source overview rows

ramteid:
## Summary
- Accept connector-manager `/sources` overview rows in `Source.from_row`.
- Preserve existing flat DB row parsing.
- Add unit coverage for both shapes.

## Why This Change Was Required
- The AI service expected flat source rows (`id`, `name`, `source_type`, ...) but connector-manager `/sources` returns `SourceSyncOverview` objects with a nested `source` payload.
- That mismatch triggered key lookup failures during source parsing, which caused source discovery to fail in chat sessions.
- When source discovery failed, the model was more likely to behave as if no apps were connected, which directly degraded RAG retrieval behavior for email/chat queries.
- Supporting both payload shapes removes this integration fragility and keeps chat source context plus retrieval paths reliable.

## Validation
- Unit tests cover both flat and wrapped source rows.
- Manual runtime verification confirmed wrapped connector-manager rows parse successfully.

### DIFF
--- a/services/ai/db/models.py
+++ b/services/ai/db/models.py
@@ -120,13 +120,12 @@ class Source:
 
     @classmethod
     def from_row(cls, row: dict) -> "Source":
-        source = row.get("source") if isinstance(row.get("source"), dict) else row
         return cls(
-            id=source["id"],
-            name=source["name"],
-            source_type=source["source_type"],
-            is_active=source["is_active"],
-            is_deleted=source["is_deleted"],
+            id=row["id"],
+            name=row["name"],
+            source_type=row["source_type"],
+            is_active=row["is_active"],
+            is_deleted=row["is_deleted"],
         )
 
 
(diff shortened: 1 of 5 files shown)


services/ai/db/models.py
```diff
@@ -120,12 +120,13 @@ class Source:
 
     @classmethod
     def from_row(cls, row: dict) -> "Source":
+        source = row.get("source") if isinstance(row.get("source"), dict) else row
```
prvnsmpth: I think it would be better to handle this wherever we parse the response from the `/sources` call to connector.

This `from_row` method should always expect a row: dict from the sources table. It should not handle any other format.

ramteid: Agreed — moved the unwrapping to both `_fetch_sources` call sites (executor.py and chat.py) where we parse the `/sources` response, and reverted `from_row` to always expect a flat source dict.

PR #275, opened 2026-06-04
fix(caddy): avoid health checks for optional connectors

ramteid:
## Summary
- Remove active health checks from Google and Atlassian reverse proxy routes.
- Keep the routes available when those optional connector profiles are enabled.
- Stop Caddy from repeatedly resolving disabled connector services.


ramteid:
on second thought, this might not be required

PR #276, opened 2026-06-04
fix(content): quietly skip invalid PDFs

ramteid:
## Summary
- Treat normal PDF parser failures as an empty extraction result instead of a warning-level extraction failure.
- Preserve panic protection for malformed PDFs that unwind inside the parser.
- Add coverage for a PDF catalog missing `/Pages`, matching the noisy log signature.

## Why
In production invalid PDFs will occur sometimes. The app must threat these cases resiliently.

### DIFF
--- a/shared/src/content_extractor.rs
+++ b/shared/src/content_extractor.rs
@@ -153,7 +153,7 @@ fn extract_pdf_text(data: &[u8]) -> Result<String> {
     match result {
         Ok(Ok(text)) => Ok(text.trim().to_string()),
         Ok(Err(e)) => {
-            debug!("Skipping PDF with unextractable text: {}", e);
+            warn!("Skipping PDF with unextractable text: {}", e);
             Ok(String::new())
         }
         Err(_) => {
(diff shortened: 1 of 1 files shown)


shared/src/content_extractor.rs
```diff
@@ -152,7 +152,10 @@ fn extract_pdf_text(data: &[u8]) -> Result<String> {
 
     match result {
         Ok(Ok(text)) => Ok(text.trim().to_string()),
-        Ok(Err(e)) => Err(anyhow!("Failed to extract text from PDF: {}", e)),
+        Ok(Err(e)) => {
+            debug!("Skipping PDF with unextractable text: {}", e);
```
prvnsmpth: Can we log at `warn` instead? Would be nice to flag this. Otherwise users will see no indication that certain files were not indexed. Ideally we should also have a metric counter but can come later.

ramteid: Good point — changed to `warn!`.

PR #281, opened 2026-06-04
fix(chat): clean up stream state on stop and on conversation switch

ramteid:
## Problem

Two related bugs in the chat EventSource lifecycle caused the input to become unresponsive.

**Stop button leaves input unusable.** `handleStop()` closed the EventSource but did not stop the thinking-text rotation interval, reset `activeStreamingMessageId`, or call `refreshProcessedMessages()`, leaving the UI in a half-stopped state. Additionally, some browsers fire the EventSource `error` event asynchronously after a manual `close()` call. `handleConnectionError` treated this as a genuine connection failure, set an error message, and performed redundant cleanup — even though the stop was intentional. As a result, the input box appeared stuck even though `isStreaming` was false.

**Switching between conversations leaves the old stream running.** `afterNavigate()` reloaded the message list and reset branch/edit state, but did not close the active EventSource from the previous chat. The old stream continued firing event handlers that mutated `chatMessages` for the wrong chat, making the newly loaded conversation appear blank or unresponsive.

## Changes

**`web/src/routes/(app)/chat/[chatId]/+page.svelte`**

`handleStop()`:
- Now calls `stopThinkingText()` to cancel the thinking-text rotation timers.
- Resets `activeStreamingMessageId` to `null` and calls `refreshProcessedMessages()` so the message list reflects the stopped state immediately.
- Clears the `error` state so no stale error banner is shown after an intentional stop.

`handleConnectionError()`:
- Added an early return when `!isStreaming`. `handleStop()` sets `isStreaming = false` synchronously before the EventSource fires its `error` event, so this guard reliably distinguishes an intentional stop from a genuine connection failure.

`afterNavigate()`:
- Now closes and nulls the active `eventSource` before applying the new chat's data.
- Resets `isStreaming`, `error`, and the thinking-text timers so the new conversation starts from a clean state.

### DIFF
--- a/web/src/routes/(app)/chat/[chatId]/+page.svelte
+++ b/web/src/routes/(app)/chat/[chatId]/+page.svelte
@@ -82,6 +82,14 @@
     })
 
     afterNavigate(() => {
+        // Tear down any active stream from the previous chat before loading the new one.
+        if (eventSource) {
+            eventSource.close()
+            eventSource = null
+        }
+        isStreaming = false
+        error = null
+        stopThinkingText()
         chatMessages = [...data.messages]
         branchSelections = {}
         activeStreamingMessageId = null
@@ -381,7 +389,12 @@
             eventSource.close()
             eventSource = null
         }
+        // Reset all stream state so the input is immediately ready for a new message.
         isStreaming = false
+        error = null
+        activeStreamingMessageId = null
+        stopThinkingText()
+        refreshProcessedMessages()
         requestAnimationFrame(() => recalcBottomPadding())
         userInputRef?.focus()
     }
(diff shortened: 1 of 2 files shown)


web/src/routes/(app)/chat/[chatId]/+page.svelte
```diff
@@ -82,6 +82,14 @@
     })
 
     afterNavigate(() => {
+        // Tear down any active stream from the previous chat before loading the new one.
```
Comment (prvnsmpth) [inline web/src/routes/(app)/chat/[chatId]/+page.svelte:85]: I've added a new e2e test suite for the chat streaming, maybe we could add a couple tests there to cover this behaviour? See `web/e2e/chat-stream.spec.ts`

Comment (ramteid) [reply web/src/routes/(app)/chat/[chatId]/+page.svelte:85]: Added two tests to `web/e2e/chat-stream.spec.ts`: one that clicks the stop button mid-stream and asserts the input is immediately ready for a new message, and one that navigates to a different chat while streaming and asserts the new chat shows no streaming state.

PR #284, opened 2026-06-05
fix(ai): prefer indexed text over binary fetch for office documents

ramteid:
## Problem

When the AI agent tried to read a Nextcloud DOCX file from chat (e.g. "Test.docx"), it consistently failed with:

\`\`\`
read_document error: Client error '400 Bad Request' for url 'http://connector-manager:3004/action'
\`\`\`

The agent found the document via search, identified its ID, and called \`read_document\` — but instead of returning the contract text it got a 400 error and gave up. The root cause: \`read_document\` classifies DOCX as a binary type and always calls \`fetch_file\` on the connector, even though the full document text had already been extracted and stored in the database at index time.

## Root cause

\`document_handler.py\` has a \`BINARY_CONTENT_TYPES\` set that includes \`application/vnd.openxmlformats-officedocument.wordprocessingml.document\`. When the AI calls \`read_document\` on a DOCX the code unconditionally takes the \`_fetch_binary\` path and never checks whether indexed text is already available via \`content_id\`.

The DB confirmed both problem files had full text stored:

| Document | content_id | size |
|---|---|---|
| Test.docx | 01KT78KMXCZ6K1279H99RFQZ9F | 11 560 bytes |
| Test.docx (copy) | 01KT84W0ZMKQQKT99HV4270YG4 | 11 563 bytes |

## Fix

Before calling \`_fetch_binary\`, check the size of the existing content blob via \`get_metadata()\` (a single lightweight DB query — no full content read). If \`size_bytes > 800\`, the blob contains real document text and is returned directly via \`_read_text\`. Below that threshold the blob is a metadata-only header (images, videos, unextractable archives) — those still go through binary fetch.

The 800-byte threshold was validated against the production database: metadata-only blobs peak at ~706 bytes across all content types; real document blobs start above that. No MIME-type lists needed.

The companion connector fix (implementing \`fetch_file\` on the Nextcloud connector so the binary path works as a fallback) shipped in #279.

## Test plan

- [ ] Open a Nextcloud DOCX file in chat → document text returned without error
- [ ] Open a Nextcloud image file in chat → binary fetch path still used
- [ ] Open a Google Docs file in chat → indexed text returned directly
- [ ] \`get_metadata\` not called for non-binary types (PDFs, plain text)

### DIFF
--- a/services/ai/tools/document_handler.py
+++ b/services/ai/tools/document_handler.py
@@ -167,7 +167,11 @@ class DocumentToolHandler:
                     meta = await self._content_storage.get_metadata(doc.content_id)
                     has_extracted_text = meta.size_bytes > 800
                 except Exception:
-                    pass  # can't determine — fall through to binary fetch
+                    logger.debug(
+                        "get_metadata failed for content_id %s, falling back to binary fetch",
+                        doc.content_id,
+                        exc_info=True,
+                    )
 
             if is_binary and not has_extracted_text and self._connector_manager_url and doc.source_id:
                 return await self._fetch_binary(doc, document_name, context)
--- a/services/ai/tools/document_handler.py
+++ b/services/ai/tools/document_handler.py
@@ -60,7 +60,7 @@ DOCUMENT_TOOL = {
         "properties": {
             "id": {
                 "type": "string",
-                "description": "The document ID (from search results)",
+                "description": "The document ID. Use the [_ref:ULID] value from search results. Never pass a filename or URL.",
             },
             "name": {
                 "type": "string",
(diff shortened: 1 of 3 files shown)


services/ai/tools/document_handler.py
```diff
@@ -155,7 +155,25 @@ async def execute(
...
+            # store it in content_blobs. When that happens, prefer the stored
+            # text over a round-trip binary fetch. We detect "real content" by
+            # blob size: metadata-only blobs (written for images, videos, and
+            # unextractable archives) top out at ~706 bytes in practice; any blob
+            # above this threshold contains actual document text.
+            has_extracted_text = False
+            if is_binary and doc.content_id and self._content_storage:
+                try:
+                    meta = await self._content_storage.get_metadata(doc.content_id)
+                    has_extracted_text = meta.size_bytes > 800
```
prvnsmpth: are we sure about this bytes > 800 check? Seems kinda brittle to me. I think a more robust way to check this would be to check the mime type of the doc if available? WDYT

ramteid: Good catch! Replaced with MIME type check: meta.content_type.startswith('text/'). All indexed content blobs use text/plain; binary-only blobs keep their original MIME type. Fixed in 455feb09.

PR #287, opened 2026-06-05
fix(ai): cite sources by URL instead of exposing internal document IDs

ramteid:
## Problem

When the AI cited a Nextcloud document in its response, it was generating text like:

```
Concept contents.pdf
Nextcloud ID: 01KT8XWK40MNCFMC9V1W2G0AP8
```

instead of a clickable link to the document. The `01KT8...` value is an internal Omni document ID (ULID) — meaningless to users and not a Nextcloud identifier at all.

## Root cause

Each `SearchResultBlockParam` includes metadata text blocks. The first block was labelled `[Document ID: <ULID>]`. When Claude followed the system prompt instruction to "always reference the source document", it read this block together with `[Source: nextcloud]` and combined them into "Nextcloud ID: <ULID>" rather than using the document's actual URL.

## Fix

**`services/ai/tools/search_handler.py`** — rename the internal document ID block from `[Document ID: ...]` to `[doc_ref: ...]`. The ULID is still available for `read_document` and `search_documents(document_id=...)` calls; it is just no longer labelled in a way that invites Claude to surface it as a user-visible citation.

**`services/ai/prompts.py`** — replace the generic "always reference the source document" citation instruction with an explicit rule: create a markdown link `[Document Name](URL)` using the `[URL:...]` field from the search result, and never expose `doc_ref` values or internal IDs to the user.

With these two changes, citations will be rendered as clickable filename links pointing to the original document (e.g. the Nextcloud `/f/<file_id>` URL).


services/ai/tools/search_handler.py
```diff
@@ -288,14 +288,17 @@ async def _execute_search(
             ]
 
             metadata_blocks = [
-                TextBlockParam(type="text", text=f"[Document ID: {doc.id}]"),
+                TextBlockParam(type="text", text=f"[doc_ref: {doc.id}]"),
```
prvnsmpth: Can we also check if this impacts the `read_document` and other tools that need to use the internal document ID? Would be nice to make sure that models (can test with any of the frontier models) are able to understand that this doc_ref can be passed as an arg to read_document, or any other tools that require a document ID.

ramteid: Good point — addressed in 0439383f. Updated `read_document`'s description and its `id` param to explicitly say "use the doc_ref value from search_documents results", and updated `search_documents`'s `document_id` param description to reference doc_ref as well. The model now has a direct textual mapping between the two tools.

prvnsmpth:
@ramteid I think this one needs merging as well

ramteid:
> @ramteid I think this one needs merging as well

Done

PR #290, opened 2026-06-05
feat(imap,nextcloud): show human-readable email/file locations instead of raw IDs

ramteid:
## Problem

IMAP email sources in chat responses showed raw document IDs (e.g. `IMAP IDs: 01KT6XNKDND5DBWKGVCQ41J49P / ...`) because no URL was available and the citation source fell back to `<unknown>`.

Nextcloud citations referenced documents via opaque internal ULIDs (`[Document ID: 01JXXXXX]`) in the LLM context, causing non-Anthropic LLMs to emit those IDs in response text rather than meaningful locations.

### IMAP

The IMAP connector now emits a `doc_source` document attribute encoding the full email location:

```
imap:{account} / {folder} / {YYYY-MM-DD} / {subject}
```

Subject is omitted when it is the placeholder sentinel. The `imap:` prefix allows generic detection without connector-specific logic in core services.

The chat UI displays this as:
- **Subject** (bold title line)
- **account · folder · date** (muted subtitle)
- Mail icon; no broken URL

### Nextcloud

- The connector now emits a `path` attribute (e.g. `/Documents/report.pdf`) so it is available in search results and to the LLM context
- Search result cards show the relative path as a subtitle below the filename
- Chat citations show the Nextcloud web URL as a clickable link

### Generic improvement (all connectors)

The `[Document source: ...]` block shown to the LLM uses, in priority order:
1. `doc_source` attribute (IMAP location string)
2. Document URL (Nextcloud, Google Drive, Jira, …)
3. Opaque ULID (last resort only)

The `[URL: ...]` metadata block is only appended when a URL exists.

## Changed files

| File | Change |
|---|---|
| `connectors/imap/src/models.rs` | Emit `doc_source` attribute: account / folder / date / subject |
| `connectors/nextcloud/src/sync.rs` | Emit `path` attribute (relative path) |
| `services/ai/tools/search_handler.py` | Generic doc_source/URL priority for LLM document identifier |
| `web/src/lib/components/search-results/imap-citation-source.svelte` | New component: renders IMAP email location as subject + account·folder·date |
| `web/src/lib/components/search-results/search-result-item.svelte` | IMAP + Nextcloud metadata subtitles |
| `web/src/routes/(app)/chat/[chatId]/+page.svelte` | Uses ImapCitationSource component; removes inline parseImapEmailSource() |
| `web/src/lib/utils/icons.ts` | Comment update for 4-part imap: format |


services/ai/tools/search_handler.py
```diff
@@ -216,6 +215,47 @@ def _build_search_tools(
...
+def _build_imap_source(doc) -> str:
+    """Build a human-readable source label for IMAP email documents.
+
+    Format: imap:{account} / {folder} / {date}
+
+    The "imap:" prefix allows the chat UI to detect and parse these sources
+    without resorting to a pseudo-URL scheme that would expose URL-encoded
+    characters verbatim in the Anthropic 【source】 citation markers.
+    """
+    extra = (doc.metadata or {}).get("extra") or {}
```
prvnsmpth: We should avoid having any connector-specific logic in core omni services. 

For this particular case, I think we can keep the logic contained in the imap connector code. We can emit events with this `doc_source` attribute. The search tool will automatically render all attributes in the search results.

web/src/routes/(app)/chat/[chatId]/+page.svelte
```diff
@@ -192,6 +193,26 @@
         return arr[Math.floor(Math.random() * arr.length)]
     }
 
+    /**
```
Comment (prvnsmpth) [inline web/src/routes/(app)/chat/[chatId]/+page.svelte]: Same as above, but with an exception. We do have connector-specific stuff in the web code, but it's kinda of contained in specific packages.

I think we can put this imap-specific stuff in `web/src/lib/components/search-results`. WDYT?

ramteid (commented):
There was a similar issue for Nextcloud sources - added the fix to this pr

services/ai/tools/search_handler.py
ramteid: All IMAP-specific logic has been moved out of `search_handler.py` into `connectors/imap/src/models.rs`. The connector now emits a `doc_source` attribute on every indexed email thread with format `imap:{account} / {folder} / {YYYY-MM-DD} / {subject}`. The search tool reads it generically via `(doc.attributes or {}).get("doc_source")` — no connector-specific logic remains in core AI services.

web/src/routes/(app)/chat/[chatId]/+page.svelte
Comment (ramteid) [reply web/src/routes/(app)/chat/[chatId]/+page.svelte]: The inline `parseImapEmailSource()` function and location rendering have been extracted into a new dedicated component: `web/src/lib/components/search-results/imap-citation-source.svelte`. Both the chat citations (`+page.svelte`) and search result cards (`search-result-item.svelte`) now import and use this component.

prvnsmpth:
@ramteid seems like there's some code changes conflicting with #284 which I already merged

ramteid:
> @ramteid seems like there's some code changes conflicting with #284 which I already merged

Right, resolved conflicts now.

PR #324, opened 2026-06-15
fix(connector-manager): strip connector_state and checkpoint from /sources API response

ramteid:
## Problem

Visiting `/admin/settings/integrations` caused multiple compounding failures:

1. **Node.js heap OOM crash** — `GET /sources` on the connector manager returned ~127 MB. IMAP `connector_state` fields (42 MB + 15 MB) and `SyncRun.checkpoint` fields (970 KB × 10 runs for Nextcloud) were serialized verbatim into the API response, exhausting the SvelteKit SSR heap.

2. **6–10 second page load** — Even after fixing the response size, the endpoint still took 4+ seconds because `find_all_sources()` selected `connector_state` and `checkpoint` from the sources table. PostgreSQL decompresses TOAST values before transfer; sqlx then JSON-parses them into `serde_json::Value`. This transferred and parsed ~116 MB on every request.

3. **IMAP connectors permanently unhealthy** — The connector manager embeds `source.checkpoint` (42 MB) in the `SyncRequest` body it POSTs to each connector's `/sync` endpoint. Axum's default 2 MB body limit in the SDK server rejected every IMAP sync with HTTP 413.

4. **120 MB Postgres → connector-manager transfer per request** — `list_runs_for_sync_types` used `SELECT sr.*` inside a window-function subquery, pulling `checkpoint` (~970 KB/run × 700 runs) even though neither the health evaluator nor the response uses it.

5. **Root cause of the IMAP checkpoint bloat** — the IMAP connector was using `connector_state` as a message-body cache: each indexed message's fully-processed body (inline text + extracted attachment text) was persisted in `FolderSyncState.messages`, which is what produced the 42 MB checkpoints in (1)/(3) on large mailboxes.

## Fix

| Location | Change |
|---|---|
| `shared/src/db/repositories/source.rs` | Add `find_all_sources_without_state()` — returns `NULL::jsonb` for `connector_state` and `checkpoint`, avoiding TOAST decompression |
| `services/connector-manager/src/handlers.rs` | `list_sources` uses the new lightweight query; `build_source_sync_overviews` strips `connector_state`/`checkpoint` from both `Source` and `SyncRun` before serializing |
| `shared/src/db/repositories/sync_run.rs` | `list_runs_for_sync_types` replaces `SELECT sr.*` with an explicit column list and `NULL::jsonb AS checkpoint` |
| `sdk/rust/src/server.rs` | Add `DefaultBodyLimit::disable()` to the connector SDK router — consistent with the connector manager's own server |
| `connectors/imap/src/models.rs`, `connectors/imap/src/sync.rs` | Stop persisting message bodies in the checkpoint at all (root-cause fix for #1/#3, not just a symptom workaround): `ParsedEmail.body_text`/`body_is_html` are now `#[serde(skip)]`, so `connector_state` only stores lightweight per-message metadata. When a thread document needs rebuilding (new message in the thread, a flag change, or a deletion), the bodies of the thread's other messages are re-fetched from the IMAP server via a new `ensure_bodies_loaded`, with a per-sync-run cache so each message is fetched at most once. A connection error during a re-fetch aborts the folder (retried next run); a single message the server won't return degrades to headers-only and self-heals later. |

## Test plan

- [ ] Visit `/admin/settings/integrations`
- [ ] `GET /sources` response is under 20 KB
- [ ] IMAP sync runs to completion and existing oversized checkpoints shrink on the next save
- [ ] A thread with a new reply, a flag change, or a deletion still rebuilds with full body content for all thread members

### DIFF
--- a/connectors/imap/src/models.rs
+++ b/connectors/imap/src/models.rs
@@ -27,7 +27,10 @@ pub struct FolderSyncState {
     /// for deletion detection (`indexed_uids − server_uids`).
     #[serde(default)]
     pub indexed_uids: Vec<u32>,
-    /// Parsed message snapshots keyed by IMAP UID, used to rebuild thread docs.
+    /// Per-message metadata snapshots keyed by IMAP UID, used to rebuild thread
+    /// documents.  The message body (`ParsedEmail::body_text`) is intentionally
+    /// **not** persisted here; it is re-fetched from the IMAP server on demand
+    /// when a thread is rebuilt (see `SyncManager`'s `ensure_bodies_loaded`).
     #[serde(default)]
     pub messages: HashMap<u32, ParsedEmail>,
     /// UIDs that were intentionally skipped due to the `max_message_size` limit.
@@ -65,10 +68,18 @@ pub struct ParsedEmail {
     pub to: Vec<String>,
     pub cc: Vec<String>,
     pub date: Option<OffsetDateTime>,
+    /// The indexable message body.  Deliberately **not** persisted in the
+    /// connector checkpoint (`#[serde(skip)]`): bodies — especially once
+    /// extracted attachment text is appended — can be megabytes each, and
+    /// persisting one per indexed message blew `Source.connector_state` up to
+    /// >100 MB on large mailboxes.  Only lightweight metadata is checkpointed;
+    /// the body is re-fetched from the server when a thread document is rebuilt
+    /// (see `SyncManager`'s `ensure_bodies_loaded`).
+    #[serde(skip)]
     pub body_text: String,
     /// When true, `body_text` is raw HTML that must be converted via the
-    /// connector manager before indexing.
-    #[serde(default)]
+    /// connector manager before indexing.  Not persisted — see `body_text`.
+    #[serde(skip)]
     pub body_is_html: bool,
     #[serde(default)]
     pub flags: Vec<String>,
--- a/connectors/imap/src/sync.rs
+++ b/connectors/imap/src/sync.rs
@@ -35,6 +35,150 @@ fn is_connection_error(e: &anyhow::Error) -> bool {
     })
 }
 
+/// Run the HTML→text conversion and attachment-text extraction that turn a
+/// freshly parsed message into its final indexable body.  Shared by the
+/// new-message indexing pass and the on-demand body re-fetch used when
+/// rebuilding a thread document, so a re-fetched body is byte-for-byte
+/// identical to the one produced at first index time.
+async fn finalize_email_content(
+    ctx: &SyncContext,
+    sync_run_id: &str,
+    email: &mut ParsedEmail,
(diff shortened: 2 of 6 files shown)


prvnsmpth:
@ramteid Thanks for the fix! Both `connector_state` and `checkpoint` are intended to be very lightweight json objects to persist connector-level and sync-level state.

I just checked the imap connector and found that we're storing the entire message body in the checkpoint, and that's why the size blows up. It is using the checkpoint store as a message cache, but ideally we should only be storing metadata here.

We can merge this PR, but I think we should also look at optimizing checkpointing in the imap connector. A few options come to mind:

Option 1: Index one doc per email message (currently we map one entire email thread -> document in Omni)
Option 2: Keep the current email thread -> document logic, but re-fetch from source when rebuilding an entire thread. Use an in-memory cache to avoid re-fetching the same email message multiple times.

Let me know if any of these sound reasonable.

ramteid:
Thanks @prvnsmpth — agreed the real issue is that the checkpoint was being used as a message cache. I went with **Option 2** and pushed it in 1b33702c.

**Why Option 2 over Option 1:** keeping the thread→document model avoids changing `document_id`s, re-indexing every existing email thread, and reworking citations/dedup, and it matches your "keep the current logic, re-fetch on rebuild" suggestion. Option 1 (one doc per message) is a bigger, more invasive change we can still revisit later if per-message docs are desirable for other reasons.

**What the commit does:**
- `ParsedEmail.body_text` / `body_is_html` are now `#[serde(skip)]`, so `connector_state` stores only lightweight metadata (message-id, threading headers, flags, dates, etc.). Existing bloated checkpoints shrink automatically on the next save — the old body fields are simply ignored on load, so no migration is needed.
- When a thread document is rebuilt (a new message in an existing thread, a flag change, or a deletion), the bodies of the thread's *other* members are re-fetched and re-processed from the server via a new `ensure_bodies_loaded`. A per-sync-run cache (`bodies_loaded`) guarantees each message is fetched at most once no matter how many rebuilds reference it — that's the in-memory cache you described.
- The HTML→text + attachment-extraction logic is factored into `finalize_email_content`, shared by the first-index and re-fetch paths so a re-fetched body is identical to the one produced at first index time.

**Behavior / trade-offs:**
- Memory also improves: instead of loading every body from the checkpoint, only bodies for threads actually touched this run are held in memory.
- A connection error during a re-fetch aborts the folder (it reconnects on the next run) rather than emitting a document with missing content; an individual message the server won't return degrades to headers-only and self-heals on a later sync.
- The cost is extra IMAP `FETCH`es proportional to the size of the threads touched in a run, bounded by the cache. Single-message threads — the common case — incur no extra fetches.

I wasn't able to run this against a live IMAP server in my environment, so a second look at `ensure_bodies_loaded` and the three rebuild call sites in `sync.rs` would be appreciated. Happy to keep this in the same PR alongside the `/sources` fix, or split it into its own — whatever you prefer.

PR #326, opened 2026-06-16
fix(connectors): hide nextcloud/imap actions from all chat tools

ramteid:
## Problem

Nextcloud and IMAP authenticate with a single **org-level basic-auth credential** and have **no per-user OAuth flow**. Their connector actions (`nextcloud__fetch_file`, `imap__validate_credentials`, `imap__list_folders`) were advertised as chat tools.

When the chat model invoked one, the call ran user-scoped and dead-ended in the connector-manager credential gate: `resolve_credentials` found no per-user credential, saw the org row, and returned `NeedsUserAuth` → the UI rendered a **"Connection required"** card that can never be satisfied (there is no OAuth app client for these sources). The model wasted turns trying to "open" files whose content was already in context via search/`read_document`.

## Fix

Added a dedicated `ActionDefinition.hidden: bool` field (`shared/src/models.rs`) that excludes an action from **every** chat/agent tool surface unconditionally, admins included — checked in both chat-tool-list-building surfaces, the AI service's `connector_handler` (`_build_tools`) and the connector-manager `list_actions` endpoint.

This replaces an earlier approach (pointing `source_types` at a sentinel type the connector never serves) that a reviewer correctly flagged as fragile and indirect. `hidden` is a separate concern from `admin_only`, which keeps its existing meaning — visible to admins in chat, hidden from non-admins — and its separate role in dispatch-time credential resolution (org vs per-user) in connector-manager. The two concerns (chat visibility vs. credential routing) are now cleanly orthogonal booleans rather than overloading one field.

The actions stay in the manifest with `Read` mode and real `source_types`, so connector-manager still **dispatches them by name** for setup (`validate_credentials`) and the internal `read_document` binary fetch (`fetch_file`), and passes the read-only guard.

Nextcloud `fetch_file` is additionally marked `admin_only: true` so `/action` dispatch resolves the **org credential**, keeping `read_document`'s internal binary fetch working for org-level agents and non-admin users.

## Scope

- `shared/src/models.rs`: new `ActionDefinition.hidden` field
- `connectors/nextcloud/src/connector.rs`: `fetch_file` → `hidden: true`
- `connectors/imap/src/connector.rs`: `validate_credentials`, `list_folders` → `hidden: true`
- `connectors/google/src/connector.rs`, `connectors/atlassian/src/connector.rs`, `connectors/filesystem/src/connector.rs`, `sdk/rust/src/mcp_adapter.rs`: add `hidden: false` to existing `ActionDefinition` literals (new required field)
- `sdk/python/omni_connector/models.py`: `ActionDefinition.hidden: bool = False`
- `services/ai/tools/connector_handler.py`: `ConnectorAction.hidden`, checked first (and unconditionally) in `_build_tools`
- `services/connector-manager/src/handlers.rs`: `list_actions` excludes `hidden` actions from the response

## Notes

- Hides the actions for all chat users, including admins.
- Takes effect once connectors are re-registered (new manifest) and the per-user `actions:{user_id}` Redis cache expires.

### DIFF
--- a/connectors/imap/src/connector.rs
+++ b/connectors/imap/src/connector.rs
@@ -92,16 +92,12 @@ impl Connector for ImapConnector {
                     },
                     "required": ["host"]
                 }),
-                // Hide this action from EVERY chat/MCP tool surface, admins included.
-                // The AI chat tool builder (connector_handler) and the
-                // connector-manager `list_actions` endpoint only surface an action
-                // when the source's type is listed in `source_types`. A sentinel type
-                // this connector never serves (any non-IMAP variant) drops it from
-                // every tool list for all users, before `admin_only` is consulted.
-                // It stays in the manifest (Read mode) and is still dispatchable by
-                // name; these are setup/util actions with no chat-facing use.
-                source_types: vec![SourceType::Linear],
+                // Setup/util action with no chat-facing use: hidden from every
+                // chat/MCP tool surface, admins included, but still in the
+                // manifest (Read mode) and dispatchable by name.
+                source_types: Vec::new(),
                 admin_only: false,
+                hidden: true,
             },
             ActionDefinition {
                 name: "list_folders".to_string(),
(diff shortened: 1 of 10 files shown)


connectors/imap/src/connector.rs
```diff
@@ -92,7 +92,15 @@ impl Connector for ImapConnector {
...
-                source_types: Vec::new(),
+                // Hide this action from EVERY chat/MCP tool surface, admins included.
+                // The AI chat tool builder (connector_handler) and the
+                // connector-manager `list_actions` endpoint only surface an action
+                // when the source's type is listed in `source_types`. A sentinel type
+                // this connector never serves (any non-IMAP variant) drops it from
+                // every tool list for all users, before `admin_only` is consulted.
+                // It stays in the manifest (Read mode) and is still dispatchable by
+                // name; these are setup/util actions with no chat-facing use.
+                source_types: vec![SourceType::Linear],
```
prvnsmpth: Why not just mark it admin_only? That way it won't be available in chat or to any agent.

In any case, using a random non-imap connector here to mask it is not the right way to address this. Maybe we should have an enum `visibility` field that determines where this action should be made available. It can have values like 'all', 'hidden', 'admin' etc. depending on what levels we want to support.

ramteid: Good point — pushed 0447455b. Added a dedicated `ActionDefinition.hidden` field instead of the source_types sentinel. `hidden` excludes the action from every chat/agent tool surface unconditionally (admins included), while `admin_only` keeps its existing meaning (visible to admins, hidden from non-admins) plus its separate role in dispatch-time credential resolution in connector-manager — so `admin_only: true` alone wasn't quite equivalent to fully hidden, since an admin user/agent would still see it in chat. Didn't go all the way to a multi-value `visibility` enum since the two concerns (chat visibility vs. credential routing) are now cleanly separated as two booleans; happy to revisit if you'd still prefer the enum.

PR #327, opened 2026-06-16
feat(chat): decouple LLM generation from browser connection via Redis stream

ramteid:
## Summary

- **Producer/Consumer split**: LLM generation runs as a background asyncio task writing events to a Redis Stream (`XADD`). The HTTP consumer tails the stream by offset, so browser disconnects (tab switch, minimize, network blip) no longer abort generation.
- **Resumable streams**: Every SSE event carries a Redis stream ID as `id:`. Native `EventSource` reconnects and manual reconnects both forward `Last-Event-ID` / `?last_event_id` to resume from the last seen offset with no duplicate events.
- **Explicit Stop**: New `POST /api/chat/[chatId]/stop` (web) → `POST /chat/{id}/cancel` (AI) sets a Redis cancel flag; the producer checks it at iteration boundaries instead of relying on `is_disconnected()`.
- **Single DB writer**: The producer is now the sole writer of chat messages (including `content_text`), eliminating any idempotency/upsert concern on reconnect.
- **Client resilience**: Exponential-backoff reconnect on error (up to 6 attempts), `visibilitychange` reconnect when tab returns to foreground, 20 s stall watchdog, and a Stop button that terminates server-side generation.
- **No duplicate runs on reconnect**: a follow-up commit closes a race where reconnecting after the stream's TTL had expired could fall through and start a second generation; the consumer now returns an explicit `not_resumable` event instead.
- **Lock kept alive during silent gaps**: the per-chat run lock was only refreshed when an SSE event was produced, so a long silent stretch in the agent loop (e.g. a slow tool call with no intermediate output) could let the lock expire while generation was still running, opening a window for a duplicate producer. A periodic 60 s heartbeat now refreshes the lock independently of event production.

## Test plan

- [ ] Start a long generation, switch tabs for >30 s, return — stream resumes seamlessly
- [ ] Start a generation, minimize the browser, restore — no duplicate messages, continues where it left off
- [ ] Reload mid-stream — page reconnects with last offset and renders remaining tokens
- [ ] Click Stop — generation ends server-side (no further tool calls or tokens)
- [ ] Two rapid reconnects — only one producer runs (lock prevents duplicate run)
- [ ] A generation with a long silent gap (slow tool call) past the old 300 s lock TTL — no duplicate producer spawned
- [ ] `npm run check` (web TypeScript) passes

### DIFF
--- a/services/ai/db/messages.py
+++ b/services/ai/db/messages.py
@@ -9,6 +9,25 @@ from .models import ChatMessage
 from .connection import get_db_pool
 
 
+def _extract_content_text(message: Dict[str, Any]) -> Optional[str]:
+    """Plain-text projection of a message for search/title. Mirrors omni-web's
+    extractContentText so rows written here match those written by the web."""
+    if message.get("role") not in ("user", "assistant"):
+        return None
+    content = message.get("content")
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list):
+        parts = [
+            b.get("text", "")
+            for b in content
+            if isinstance(b, dict) and b.get("type") == "text"
+        ]
+        joined = "\n".join(p for p in parts if p)
+        return joined or None
+    return None
+
+
 class MessagesRepository:
     def __init__(self, pool: Optional[Pool] = None):
         self.pool = pool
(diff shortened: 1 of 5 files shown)


prvnsmpth:
Thanks for the fix! Will review this soon.

PR #286, opened 2026-06-05
fix(ui): improve suggestion chips mobile layout and prompt quality

ramteid:
## Summary

- Fix horizontal overflow on mobile: suggestion buttons now use `w-full` + `text-left` instead of `truncate` + `max-w-screen-md`, preventing horizontal scrolling
- Improve AI prompt for suggested questions: rewritten to generate practical workplace search queries (e.g. "How do we handle customer refunds?") instead of literal questions about file metadata like URLs, properties tables, or git objects

PR #328, opened 2026-06-18
fix(searcher): improve home-screen suggestion quality

ramteid:
## Problem

The "Try asking" suggestion chips on the home screen (new chat) sometimes surface nonsensical keyword phrases instead of real questions — e.g. **"Git repository object storage documentation"**. The follow-up to #286 improved phrasing in general, but two gaps remained: technical/code documents are still eligible for selection, and keyword-style output is never rejected.

## Root cause

`SuggestedQuestionsGenerator` picks **completely random** documents (`fetch_random_documents`, `ORDER BY RANDOM()`) across *all* indexed sources, sends a 2000-char excerpt to the model, and only validates that the result is non-empty. When the random pick lands on a technical document (e.g. a GitHub repo), the model dutifully produces a topic keyword phrase rather than a question, and nothing filters it out.

## Changes

1. **Exclude code/technical sources from selection.** `fetch_random_documents` now takes an `excluded_source_types: &[SourceType]` list; the generator excludes `Github`. The exclusion is a `NOT EXISTS` sub-filter, so an empty list excludes nothing.

2. **Sharper prompt.** It now requires a *natural question* and explicitly forbids noun-phrase/keyword output. For documents that aren't a good basis for a business question (source code, config, logs, CI output, auto-generated/technical docs), the model is told to return the sentinel `SKIP`.

3. **Output validation.** The generated text is rejected when it is `SKIP` or when it isn't a question (no `?`). A rejected document is skipped and the loop fetches another random document (existing `MAX_RETRIES` retry path).

4. **Deduplicate suggestions.** Rejected documents cause more retry rounds, and each round re-draws documents at random, so the same document — or an identical generic question from two documents — could be surfaced twice. Generation now deduplicates by document id and by question text so the chips stay distinct. On top of that application-level guard, `fetch_random_documents` takes an `excluded_document_ids` list and filters already-consumed documents in SQL (`d.id <> ALL(...)`), so each random draw returns only new documents instead of spinning on re-drawn rows — which also keeps the `num_docs_fetched < needed` termination signal accurate.

5. **Cache invalidation.** Redis cache key bumped `v1` -> `v2` so previously cached low-quality suggestions disappear immediately instead of lingering for the 24h TTL. (Also corrected the stale `// 7 days` comment on the 86400s TTL.)

## Testing notes

Per the repo's build constraints this was developed via static analysis only — no Rust build/test and no live DB were run. The one aspect that warrants a runtime sanity check is the new SQL in `fetch_random_documents` — the `NOT EXISTS` source-type sub-filter and the `d.id <> ALL($3)` exclusion, combined with the ParadeDB `@@@` permission predicate; the change is standard SQL and `@@@` already coexists with other predicates there, but it has not been executed against a real database.

Known residual limitation: the `?` guard rejects output with no question mark, but a noun phrase with a trailing `?` would still pass — the prompt is the primary defense against that, the guard is a cheap backstop.

### DIFF
--- a/services/searcher/src/suggested_questions.rs
+++ b/services/searcher/src/suggested_questions.rs
@@ -6,23 +6,37 @@ use futures_util::StreamExt;
 use redis::AsyncCommands;
 use redis::Client as RedisClient;
 use shared::utils::safe_str_slice;
-use shared::{AIClient, DatabasePool, DocumentRepository, GroupRepository, ObjectStorage};
+use shared::{
+    AIClient, DatabasePool, DocumentRepository, GroupRepository, ObjectStorage, SourceType,
+};
+use std::collections::HashSet;
 use std::sync::Arc;
 use tracing::{debug, error, info, warn};
 
-const REDIS_CACHE_KEY: &str = "suggested_questions:v1";
-const CACHE_TTL_SECONDS: u64 = 86400; // 7 days
+// Bump the version suffix whenever the generation prompt or validation changes
+// so previously cached (and now undesirable) suggestions are invalidated.
+const REDIS_CACHE_KEY: &str = "suggested_questions:v2";
+const CACHE_TTL_SECONDS: u64 = 86400; // 24 hours
 const MAX_RETRIES: usize = 5;
-const QUESTION_PROMPT_TEMPLATE: &str = r#"You are helping generate example search queries for a workplace search tool that indexes company documents, emails, and files.
+/// Source types whose documents are predominantly code or technical material and
+/// therefore make poor "Try asking" suggestions (e.g. a GitHub repo yields queries
+/// like "Git repository object storage documentation"). They are excluded from the
+/// random document selection; per-document SKIP handling in the generator still
+/// filters technical content out of the remaining sources.
+const SUGGESTION_EXCLUDED_SOURCE_TYPES: &[SourceType] = &[SourceType::Github];
+const QUESTION_PROMPT_TEMPLATE: &str = r#"You are helping generate example search queries for a workplace search tool that indexes a company's documents, emails, and files. The result is shown as a clickable "Try asking" suggestion on the home screen, so it must read like something a real employee would naturally ask.
 
-Given the following document excerpt, generate ONE example search query that an employee might realistically type into a workplace search engine to find information related to this document's topic.
+You are given an excerpt from ONE indexed document. Based on the topic it covers, write ONE natural question that a colleague might ask to find this kind of information.
 
 Rules:
-- Write the query from the perspective of someone who does NOT have the document open — they are searching for information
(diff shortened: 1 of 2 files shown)


services/searcher/src/suggested_questions.rs
```diff
@@ -6,23 +6,37 @@ use futures_util::StreamExt;
...
-- Write only the query itself, no quotes, no prefixes like "Question:" or "Query:"
+- Write a full, natural question, phrased the way a person actually speaks, and end it with a question mark. Examples: "How do we handle customer refunds?", "What are the onboarding steps for new engineers?", "Who is responsible for Q3 budget planning?"
+- Do NOT output keyword phrases, document titles, or noun phrases such as "Git repository object storage documentation". It must be a real question.
+- Ask from the perspective of someone who has NOT seen this document and is looking for the knowledge it contains.
+- Ignore document metadata and structure (URLs, file paths, IDs, timestamps, formatting, raw code, configuration values).
+
+If this document is not a good basis for a useful business question — for example it is source code, configuration, logs, build or CI output, auto-generated content, technical/system documentation, boilerplate, or has no clear business topic an employee would search for — then respond with exactly the single word SKIP and nothing else.
+
+Output only the question itself (or SKIP), with no quotes and no prefix like "Question:" or "Query:".
```
prvnsmpth, line 39: We definitely need to upgrade the suggestion system, so thanks for this PR!

But because Omni is capable of doing small tasks, running scripts on data, etc. in addition to answering questions, it would be nice to also generate suggestions like "Analyze the data in our Q4 sales report and present the data as a chart". This is not a question but an instruction to carry out some task. Maybe we can implement that as well?

Another idea for even better suggestions: instead of generating suggestions from just one document, we can gather all indexed content from the last N days (e.g., last week) and spawn an agent to explore this data. It can then come up with even better suggestions because it has a wider context to pick from.

We still don't have the capability to spawn agents to carry out arbitrary work in the background, so it might not be easy to implement in this PR, but thought I'd mention it anyway.

ramteid: Good point! Implemented the task/instruction suggestions in 841b3477.

Every third suggestion (positions 3, 6, 9) now uses a separate task prompt that produces imperative instructions like "Summarize the key takeaways from the Q3 board meeting" or "Show a chart of sales by region from the Q4 report". The two prompts share a single `generate_suggestion_from_document` helper; question validation (must contain `?`) is only applied for the question type. Cache key bumped to `v3` to evict stale question-only results.

The agent-based approach (gather recent indexed content, spawn a background agent for wider context) sounds great for a follow-up — we don't have the background agent infrastructure yet, but it would naturally slot in as a third generation strategy once that's available.

prvnsmpth: Awesome, thanks!

PR #384, opened 2026-07-27
feat(web): optional Spider Cloud crawling via SPIDER_CLOUD_API_KEY

j-mendez:
## Problem

The web connector crawls with a plain direct fetch. Sites behind Cloudflare, a WAF, or IP/geo blocking answer that fetch with 403/429/503, and [`sync.rs`](https://github.com/getomnico/omni/blob/main/connectors/web/src/sync.rs#L45) drops anything that isn't `200 OK` — so those pages silently never reach the index, with no signal that a source is only partially crawled.

## Change

Setting `SPIDER_CLOUD_API_KEY` routes crawls through [Spider Cloud](https://spider.cloud) (proxy rotation + anti-bot bypass) instead. `spider` already ships first-class support for this behind its `spider_cloud` feature, so the wiring is small.

```bash
SPIDER_CLOUD_API_KEY=sk-...      # the entire setup
SPIDER_CLOUD_MODE=smart          # optional: smart (default) | proxy | unblocker | api | fallback
SPIDER_CLOUD_API_URL=...         # optional
```

`smart` is the default: it proxies every request and automatically escalates to the unblocker API when it detects bot protection (403/429/503, Cloudflare challenges, CAPTCHA pages). An unrecognized `SPIDER_CLOUD_MODE` degrades to `smart` rather than failing a sync.

## No behavior change when unset

This is strictly opt-in, and deliberately kept that way:

- With no key, `with_spider_cloud_config` is never called, so `configuration.spider_cloud` stays `None` and every spider-side cloud branch takes the existing direct-fetch arm.
- `spider_cloud = ["serde", "reqwest_json"]` — serialization only. No Chrome, no headless browser, no new system dependencies.
- No config-schema, DB, SDK, API, or admin-UI changes. Existing stored source configs deserialize identically.
- Return format is left at `raw`. This matters: `WebPage::from_html` parses with `scraper` and `sync.rs` stores `raw_html` as `text/html`, so requesting markdown would corrupt title/description extraction and every `content_hash`, forcing a full re-index.
- The env lookup resolves once into a `OnceLock` rather than on every crawl.

## Verification

Crawled `books.toscrape.com` through `build_spider_website` in three configurations:

| Configuration | Result |
|---|---|
| No key set | 2 pages, raw HTML, titles intact — unchanged from today |
| Valid key, `smart` | 2 pages, raw HTML, titles intact |
| Invalid key, `proxy` | 0 pages — confirms the proxy is genuinely in the request path, not silently bypassed |

`cargo test -p omni-web-connector --lib` passes (11 tests, including new mode-parsing coverage) with the cloud env vars explicitly cleared. `cargo clippy` and `cargo fmt --check` are clean.

`Cargo.lock` also picks up `spider` 2.52.0 → 2.52.12, which is a net dependency reduction (drops a `toml` + `windows-sys` chain).

## Unrelated observation

`WebSourceConfig.max_pages` is parsed but never applied — `build_spider_website` never calls `.with_limit(...)`, so the configured limit is currently ignored. Left alone here to keep this diff focused; happy to open a separate issue or PR.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -170,6 +170,21 @@ GOOGLE_MAX_AGE_DAYS=712 # Documents older than this will not be indexed
 # Web Connector Configuration
 WEB_SYNC_INTERVAL_SECONDS=86400  # Daily recrawl (24 hours)
 
+# Optional: Spider Cloud (https://spider.cloud) for higher crawl success rates.
+# Sites behind Cloudflare, a WAF, or IP/geo blocking return 403/429/503 to a
+# direct fetch and are dropped from the index. Spider Cloud routes crawls
+# through rotating proxies and anti-bot bypass instead.
+# Leave SPIDER_CLOUD_API_KEY unset to keep the current direct-fetch behavior.
+# SPIDER_CLOUD_API_KEY=sk-...
+# Mode: smart (default) | proxy | unblocker | api | fallback
+#   smart     - proxy everything, auto-escalate to the unblocker on bot walls
+#   proxy     - proxy transport only, cheapest
+#   unblocker - always use the unblocker API
+#   api       - fetch each page through the crawl API
+#   fallback  - direct fetch first, cloud only after a failure
+# SPIDER_CLOUD_MODE=smart
+# SPIDER_CLOUD_API_URL=https://api.spider.cloud
+
 # Log level for all rust services
 RUST_LOG=info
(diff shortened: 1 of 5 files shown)


prvnsmpth:
Thanks for the PR! Merged

PR #400, opened 2026-08-02
feat(windshift): configure Windshift server via admin UI, drop compose env warnings

prvnsmpth:
- Windshift base URLs are now managed in the admin UI (Settings > Integrations) and stored in `connector_configs`; the connector reads them via connector-manager, with `WINDSHIFT_BASE_URL` env vars kept as a fallback for legacy deployments
- Add `SdkClient.getConnectorConfig` so TypeScript connectors can fetch admin-managed connector configs
- Default `WINDSHIFT_CONNECTOR_PORT` and `EMBEDDING_MODEL` in docker-compose so unset vars no longer emit compose WARN lines for users not running the windshift/local-embeddings services
- Remove the Windshift base URL vars from `.env.example` and update the connector README to describe the UI-based setup

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,14 +38,10 @@ GOOGLE_ADS_CONNECTOR_PORT=4016
 DARWINBOX_CONNECTOR_PORT=4017
 WINDSHIFT_CONNECTOR_PORT=4018
 
-# Public URL of the Windshift instance this Omni install talks to.
-# Reachable from the user's browser for OAuth consent. One Omni install = one
-# Windshift instance in v1. Set WINDSHIFT_INTERNAL_BASE_URL below when the
-# connector container needs a different route to that instance.
-WINDSHIFT_BASE_URL=https://windshift.example.com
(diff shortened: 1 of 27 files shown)


prvnsmpth:
@stefan-ernst Could you please review this PR and let me know if it looks ok? I've changed how the Windshift base URL is specified. The setup is now done via the UI, instead of setting env vars. Let me know if this works, thanks!

stefan-ernst:
Hey @prvnsmpth - this is good, it makes it more user friendly to switch the urls. However it introduces a problem that I lazily avoided in the first PR by making that URL only configurable via env: The URL needs to be validated against SSRF attacks. If the admin user is not the one running the local server, arbitrary services can be enumerated with the url being open in this way. In Windshift we avoided this by blocking local urls, which in turn caused a lot of issues for users, so we'll have to allow local urls by default, making it less secure out of the box.

Its a bit of a judgement call and nothing really to do with this PR in particular

prvnsmpth:
That's a good point, I'll implement SSRF validation. Thanks!

PR #423, opened 2026-08-19
feat(salesforce): add Salesforce connector with sync, permissions and actions

prvnsmpth:
- Adds a new Salesforce connector indexing Accounts, Contacts, Opportunities, Leads, Cases, and Tasks into Omni
- Implements full, incremental, and realtime sync with resumable per-object checkpointing and watermark invalidation on config changes
- Mirrors Salesforce sharing faithfully (owners, queues, role hierarchy, manual/rule shares) via person/group events
- Adds search operators, read/write actions, and admin setup UI; migration 112 registers the new source type


prvnsmpth:
### Update: per-user MCP OAuth, fail-closed credential handling, and source-scoped OAuth client management

This push completes the Salesforce MCP per-user OAuth work reviewed across two focused review rounds (verdict: ✅ ready) and adds the client-management UX requested in follow-up.

**Backend / SDK / connector-manager** (`a99ea3f2`)
- Official `sf-mcp-server` runs inside the connector container via the launcher, with a disposable, per-source/user CLI state dir; raw secrets are scrubbed from the child environment.
- MCP actions, resources, prompts, and skills require the caller's per-user OAuth credential — never the org JWT; revoked/invalid tokens surface a standardized `412 needs_user_auth`.
- Auth-failure markers are preserved by the launcher and consumed/unlinked by the SDK.
- Failed credential-ready discovery invalidates stale MCP catalogs (empty manifest, `mcp_catalog_loaded=false`) with replay on the next registration; catalogs are kept in memory only.
- URI template placeholders are restricted to single path segments; unknown non-JSON OAuth 4xx refresh failures are treated as transient instead of deleting credentials.
- Salesforce DCR metadata is recognized end-to-end with fail-closed revocation.

**Web** (`f91116ba`)
- MCP per-user OAuth client configuration moved from the connector setup dialog to the **OAuth Apps** tab, one row per connected Salesforce source (`salesforce:<source-id>`) — the shared client dialog already supports scoped Salesforce providers incl. authenticated DCR.
- Salesforce source settings page shows an **"MCP is not available yet"** warning until a usable per-user OAuth client is configured.
- DCR rotation fix: revoking an existing registered client clears its stored identity so a new initial-access token re-registers instead of fast-pathing to the revoked pair.
- Source deletion defaults to fail-closed DCR revocation with a documented `?revoke_oauth_client=false` escape.
- New focused vitest coverage for DCR revocation (19 web OAuth tests), SDK suite (24 tests), connector-manager lib tests (70).

The updated stack is built and deployed locally with all health checks green (Salesforce registered, migration 113 applied).

PR #457, opened 2026-09-14
fix(web): use node:http/https for pinned remote-MCP fetch to unbreak OAuth in container builds

stefan-ernst:
## Summary

OAuth flows against admin-configured endpoint providers (`windshift`, `remote_mcp:*`) fail in the compiled `omni-web` container with a misleading `412 — OAuth client for X is not configured`, even when everything is configured correctly. The root cause is that the SSRF-pinned fetch (`fetchWithPinnedRemoteMcpDns`) routes requests through an undici `Agent` dispatcher, and the bundled undici inside the container is broken for HTTPS.

## Root cause

The container build wraps the `undici` package in a CJS shim. Inside that shim, undici's HTTP/2 module resolution fails and is silently replaced by a stub:

```js
let http2;
try {
  http2 = require('node:http2');
} catch {
  http2 = { constants: {} };  // stub — no connect()
}
```

Any HTTPS request routed through a custom dispatcher built from this shim then fails at connect time:

```
fetch failed | cause: http2.connect is not a function
```

Because `fetchWithPinnedRemoteMcpDns` is only used for admin-configured endpoint providers, plain fetches (and all dev-mode testing) work fine — the breakage only appears in the compiled image, for exactly the OAuth flows that talk to self-hosted endpoints. We hit this connecting Omni to a self-hosted Windshift instance: dynamic client registration succeeded with an identical plain-`fetch` request from inside the same container, but Omni's own request always threw.

The failure is additionally invisible because `dynamicallyRegisterClientUnlocked` swallows all exceptions (`catch { return null }`), which is what produced the misleading 412.

## Fix

- Replace the undici dispatcher in `fetchWithPinnedRemoteMcpDns` with `node:http`/`node:https` transports. These honor a pinned `lookup` directly and remain external in the bundle. **SSRF validation and DNS-pinning semantics are preserved**: addresses are still resolved and validated up front and every request is pinned to the validated addresses via the request's `lookup` option. `redirect: 'manual'` semantics are preserved (node's client never follows redirects).
- Update `probe.test.ts` to mock the node transports instead of the `undici` module. The three network-behavior tests are unchanged in intent: bearer auth headers, session-id propagation, no redirects, DNS re-check per request, and fail-closed on DNS rebinding.
- Log dynamic-client-registration failures (previously fully swallowed) — this bug cost us an afternoon precisely because nothing was logged.

## Verification

- `vitest run src/lib/server/mcp/probe.test.ts` — 3/3 pass
- `svelte-check` error count unchanged vs. `master` (remaining errors pre-exist)
- `eslint`/`prettier` clean on the touched files
- Deployed the patched build against a live self-hosted Windshift: dynamic client registration now succeeds (was consistently failing with the `http2.connect` error before), and the full OAuth connect flow completes

## Follow-ups worth considering

- The bundled-undici shim swallowing the `node:http2` failure may break other dispatcher-based code paths in the container image; it might be worth externalizing `undici` in the bundler config or dropping the npm dependency where Node's built-in fetch suffices.
- The 412 error message ("Ask an admin to set it up under OAuth Apps") is misleading when the provider is auto-managed and registration merely failed; surfacing the logged DCR error to the UI would help self-hosters.


stefan-ernst:
Manually verified it works now with https connection after the fix

PR #458, opened 2026-09-14
feat(chat): per-chat deselection of connector sources

stefan-ernst:
Connector toolsets are advertised to the agent in every conversation, and the agent can load their tools via `tool_search` / `load_tool` / `load_tool_set` even when the source has nothing to do with the conversation. For sources that are rarely relevant this wastes turns checking those resources.

This adds a per-chat exclusion: when starting a new conversation you can deselect connected sources in the composer. Deselected sources are no longer advertised in the system prompt, `tool_search` doesn't return their tools, and `load_tool` / `load_tool_set` refuse them with a clear error. The exclusion is stored on the chat row (new `chats.excluded_source_ids` column, migration 117), so it survives reconnects and is also applied to tools restored from history. Excluded sources show up as chips on the chat page and can be re-included mid-conversation.

This is deliberately tool-scoped only — it doesn't touch the unified search index or source visibility.


prvnsmpth:
In my experience, the models are generally smart enough to load the right tools, and I've rarely seen them load tools that are not relevant to the user's request.

I'd like to keep the UX minimal as much as possible and avoid having too many controls if we can avoid it. As a user, I would expect the agent to be smart enough to load the right tools without having to explicitly select/deselect apps in the composer input.

Was this change prompted by some incorrect behaviour you noticed wrt to the tool loading? Maybe there's an alternative way to handle that issue.

stefan-ernst:
Hey, yes I noticed that the agent will spend time every turn discovering tools from Windshift for example when Windshift is connected (I believe this applies to all tools). For example you ask it "whats a good laser printer" and it will try to find the answer in Windshift :)

prvnsmpth:
Ah I see, let me think about this for a bit. The idea of restricting a chat to specific apps/sources makes sense, but I'm not entirely sure if we should have UI controls for that.

I think there might be something in the system prompt that's encouraging the model to look into the connected apps first before looking in the search index. If we can give clearer directions to the model on when it should use tools from apps vs looking up info in the index, that might be a simpler solution to the issue.

PR #460, opened 2026-09-15
feat(web): basic PWA harness

stefan-ernst:
Adds a basic PWA harness so omni can be installed to a phone's home screen: proper icon, standalone display, and app shortcuts.

The manifest, service worker, and iOS head links follow the same conservative pattern we use in Windshift: the service worker handles top-level navigations network-first with a self-contained recovery document for failed or timed-out navigations, and deliberately does not cache the app shell (a cached shell without its exact hashed asset graph produces an unbootable white page). Web Push handlers are included but inert for now — they're meant for run notifications later.

Also fixes the homescreen blank-page symptom class: launches that fail before the server (or hit a 5xx) now show a branded "Omni couldn't connect — Retry" page instead of nothing.

Draft for now — deployed and verified on my instance; marking ready after a final device check.


stefan-ernst:
Review findings addressed in the second commit:

- Real 192/512 icons (sourced from the website's 512 logo) plus separate padded maskable variants — the old 255×256 full-bleed logo is gone. apple-touch-icon regenerated at 180×180 from the high-res source with an 80% safe zone for iOS corner masking.
- Manifest `theme_color` dropped so Android falls back to the light/dark media-queried `theme-color` meta.
- The recovery document now covers network failures only — real 4xx/5xx responses pass through untouched (the 429 rate limiter was being disguised as a connectivity problem).
- Navigation timeout raised to 30s; a rejected preload response falls back to the direct fetch.
- Update flow is now consent-based: a persistent "Omni has been updated — Reload" toast, and `controllerchange` reloads only when we asked for it, so deploys no longer force-reload open tabs.
- Service worker is served from a route with `Cache-Control: no-cache` (heuristic caching could delay update detection by ~24h).
- No SW registration in dev; `notificationclick` falls back to `openWindow` when `focus()` rejects; dropped the opaque notification badge icon.

PWA_REVIEW.md
```diff
@@ -0,0 +1,88 @@
+# Mobile PWA Review — `c9e3e0e feat(web): basic PWA harness`
```
prvnsmpth: I guess we don't need to check-in this md file?

prvnsmpth (commented):
This is great! Thanks a lot for working on this.

PWA_REVIEW.md
stefan-ernst: yes, removed - this was not properly gitignored

PR #467, opened 2026-09-16
feat(microsoft): Outlook Calendar agent actions + explicit setup permission guidance

stefan-ernst:
## Summary

- **Calendar agent actions**: agents could search calendar events in the synced index, but had no actions to work with the calendar interactively. This adds `list_events` (read) and `create_event` (write) for the `outlook_calendar` source, executing with the caller's delegated token (`Calendars.Read` / `Calendars.ReadWrite`, which the connector's OAuth manifest already declares).
- **Per-user OAuth credentials crash every action (upstream master bug)**: the web OAuth callback (`web/.../oauth/callback/+server.ts`) deliberately persists refresh-fallback metadata (`client_id`, `client_secret`, `token_uri`, `token_endpoint_auth_method`, `scope`) alongside the tokens so connector-manager can refresh them — but the connector's `MSUserOAuthCreds` model used `extra="forbid"` and only accepted 3 fields, so *every* agent action executed with a connected user account failed with an unhandled pydantic `ValidationError` (HTTP 500 through connector-manager). This reproduces on a pristine master stack: connect any Microsoft account, call an action, get a 500. The model now accepts the full stored shape, and an unrecognized credential shape returns a clean 400 instead of a stack trace.
- **`execute_action` signature fix**: the SDK server calls `connector.execute_action(..., source=..., actor_email=...)`, but the Microsoft override didn't accept those kwargs, so *any* action on this connector would have raised a `TypeError` when executed through the connector-manager stack. The override now matches the base class signature, and the calendar actions use `source.config` for `graph_base_url` (same as the sync path, keeps national-cloud deployments working).
- **Setup dialog names the exact permissions and the redirect URI**: the Microsoft connect dialog said "grant both Application and Delegated permissions" without saying which ones, and never mentioned admin consent or the OAuth callback URL. Without granted consent, sources authenticate but every sync fails with Graph `Authorization_RequestDenied` ("Insufficient privileges to complete the operation"); without the redirect URI registered, per-user connect fails with `AADSTS500113`. The dialog now lists the concrete permission sets, the admin-consent step, and shows the instance's callback URL with a copy button; the connector README documents both permission tables plus the new actions.

## Notes

- `list_events` defaults to today (UTC) + 7 days and supports explicit `start`/`end` (ISO 8601, naive = UTC) and `top` (capped at 50). `create_event` normalizes times to UTC, validates attendees, and supports location, body, and Teams online meetings.
- Tested with the connector's mock-Graph harness (`tests/test_actions.py`, 9 cases incl. validation failures, the refresh-fallback credential shape, and graceful rejection of unknown credential shapes) and deployed to a live self-hosted instance against a real tenant: the registered manifest exposes both actions, and action execution reaches Graph with the caller's credential.


stefan-ernst:
Marking as draft — not ready for review yet. Live testing on a self-hosted instance surfaced an gap this PR doesn't cover: the per-user OAuth flow that powers the new actions requires the app registration to have a redirect URI registered (Entra ID → App registrations → Authentication → Web platform). With none set, the consent screen fails with `AADSTS500113: No reply address is registered for the application`. I'll fold the redirect-URI requirement into the setup dialog/README guidance alongside the permission tables before taking this out of draft.

prvnsmpth:
Thanks for this PR! Let me know if you need any help testing and merging this.

stefan-ernst:
Pushed: the setup dialog now also shows the instance's OAuth redirect URI (with copy button), and the README documents that it goes under Authentication → Web. Still in draft until I've re-verified the full per-user flow end to end.

stefan-ernst:
Fully tested with my local instance and a o365 calendar. Unfortunately I am a bit limited in testing this as I only have 1 account on my tenant and could not properly test isolation in my instance

PR #468, opened 2026-09-17
feat(chat): image attachments with paste support and vision passthrough

stefan-ernst:
## Summary

Users can already attach files to a chat via the picker or drag-and-drop, but images never reach the model as images:

- **Pasting did nothing**: the composer handled drop events but never read the clipboard, so pasting a screenshot (arguably the most natural gesture) silently failed.
- **Images were sent as generic `document` blocks**, and the AI service expands every `omni_upload` block to either inline text or a "file staged in the workspace" pointer — never an `image` content block. Vision-capable models saw a filename instead of the picture.
- **Providers dropped image blocks entirely**: the OpenAI-compatible adapter only assembles text/tool parts, and the Gemini adapter only builds text/function-call parts, so even a hand-crafted image block would be silently discarded.

This PR wires the full path:

- **Composer**: pasting clipboard files now attaches them through the same handler as drag-and-drop (chips, upload to `/api/uploads`); pure-text pastes behave exactly as before. Web-image pastes — Safari's default when copying an image on a page exposes no file, only `text/html` with a remote `` — are materialized too: `data:`/`blob:` sources client-side, remote ones through a new SSRF-guarded `POST /api/uploads/from-url` proxy (http(s) only, public-DNS validation on every redirect hop, image content-type, size/timeout caps) that returns the bytes for the normal upload flow.
- **Web**: image uploads are tagged by MIME and sent as `image` omni_upload blocks instead of `document` (message rendering, history, and the shared `OmniUploadBlock` type already allow this).
- **AI service**: at provider-call time, `omni_upload` image blocks (png/jpeg/gif/webp, ≤ 3.5 MB raw — stays under Anthropic's 5 MB request cap after base64) expand to real base64 `image` blocks — **only when the chat's model resolves as vision-capable**. Capability detection is ported from Windshift: a curated model-ID substring map plus a lenient `vision_mode: auto | on | off` override in the provider config (`on` rescues custom/local model ids, `off` force-disables), resolved on `ResolvedModel.supports_vision`. Non-vision models keep the graceful paths: the upload is staged in the workspace with a pointer text (inspectable via tools), or a short notice when no sandbox is available. Oversized or unsupported images keep the sandbox-pointer fallback too.
- **Providers**: OpenAI-compatible endpoints receive `image_url` data-URI parts, Gemini receives `inline_data` parts; Anthropic-family providers (incl. Bedrock/Vertex Claude) pass the native blocks through untouched.

## Notes

- No schema or persistence changes: `omni_upload` blocks stay compact and replayable in stored messages; expansion remains a provider-call-time concern with the existing per-call cache.
- 10 new unit tests: upload expansion (inline image, oversized fallback, unsupported media fallback, text inline, cross-user isolation) and provider conversion (image → `image_url`/`inline_data`, non-base64 sources ignored, pure-text messages unchanged).
- `svelte-check` / `pyright` clean relative to their pre-existing baselines; prettier/eslint pass on touched files.


stefan-ernst:
Tested successfully on Safari

image

services/ai/providers/gemini.py
```diff
@@ -175,6 +175,18 @@ def _convert_messages_to_gemini(
                 if document_text is not None:
                     parts.append(types.Part(text=document_text))
 
+            elif block_type == "image" and role == "user":
```
prvnsmpth, line 178: will this work for non-vision capable models? Will the image input blocks be ignored?

services/ai/providers/openai_compatible.py
```diff
@@ -177,10 +201,33 @@ def _convert_messages_to_openai(
...
             elif block["type"] == "document" and role == "user":
                 document_text = extract_text_document(block)
                 if document_text is not None:
                     text_parts.append(document_text)
+                    user_content_parts.append(
+                        ChatCompletionContentPartTextParam(
+                            type="text", text=document_text
+                        )
+                    )
+            elif block["type"] == "image" and role == "user":
```
prvnsmpth, line 219: Same as above, not sure if non-vision models will handle this gracefully across all openai-compatible providers.

Maybe we need some way to determine whether a model is vision-capable or not, not sure if there's a standard way to do this.

prvnsmpth (commented):
Thanks for this PR, this has been pending for a long time :)

services/ai/providers/openai_compatible.py
stefan-ernst: I have some detection for this in Windshift, I will port it over 😄

services/ai/providers/gemini.py
stefan-ernst: Good catch — they wouldn't have been: non-vision models would have received blocks they can't handle. Fixed in aacc2e6 by porting the capability mechanism we use in Windshift: image blocks are now only produced when the chat's model resolves as vision-capable. Non-vision models get the existing graceful paths instead — the file is staged in the workspace with a pointer text (tools like read_file can inspect it), or a short notice when no sandbox is available. Vision-capable models are unaffected.

services/ai/providers/openai_compatible.py
stefan-ernst: Agreed — there's no standard cross-provider signal, so this now uses the approach we shipped in Windshift (`internal/llm/vision_capability.go`), ported in aacc2e6:

- **Detection**: a conservative curated model-ID substring map (`gpt-4o*`, `gpt-4.1`, `o1/o3/o4-mini`, `claude-3+`, `gemini-1.5+`, `grok-3/4`, `llama-3.2/4`, `pixtral`, `*vision*`, `*-vl*`, …) — same list, kept deliberately id-based so it works uniformly across providers including OpenRouter-style namespaced ids on openai-compatible endpoints.
- **Override**: a `vision_mode: auto | on | off` key in the provider config — `on` rescues custom/local models the map can't recognize, `off` force-disables a model the map wrongly marks capable, `auto` (default) defers to the map. Missing/invalid values are leniently treated as `auto`.
- **Resolution + gating**: `ResolvedModel.supports_vision` is computed at resolve time, and `expand_uploads` only inlines base64 image blocks when it's true. A no-op default-true flag keeps the expansion safe for non-chat callers (title generation etc.).

A dedicated UI toggle can come later; the config key already works today.

stefan-ernst:
Reworked the image support quite a bit. There are a few changes to the LLM connection setup:

- The model list now renders a bit wider ans now has a vision column
- Vision support is now determined by a probe request when adding the model (tested it on a few models on fireworks)
- There is a new selector on the provider config to either probe, always send images or never send images

PR #470, opened 2026-09-19
fix(sync): let a failed source recover without a manual sync

ramteid:
## Problem

Once a source hits `SYNC_MAX_CONSECUTIVE_FAILURES`, it never syncs again on its own. The breaker blocks scheduled runs — but only a successful scheduled run can clear the failure streak. So it never reopens, and the source stays dead until someone syncs it by hand.

In one self-hosted deployment an IMAP source sat unsynced for six weeks this way, while the other sources kept syncing hourly.

How it got stuck:

1. The IMAP client has no timeouts, so an unresponsive server leaves the sync task hanging forever.
2. After an hour the manager gives up on the run and sends `/cancel`. But cancel only sets a flag that the hung task never reads, so the connector keeps holding the source's slot.
3. Every scheduled sync after that gets `409 sync already in progress`. Ten of those trip the breaker, and it latches.

## Fix

- **imap**: bound every IMAP round trip. This is a liveness bound, not a budget for the work — a folder may take as long as it needs as long as each command keeps returning. A timeout counts as a connection error, so the session is rebuilt instead of reused in an unknown protocol state, and the flag-fetch pass gives up on a dead session instead of burning one timeout per chunk. Because the bound sits inside `ImapSession`, the `validate_credentials` and `list_folders` actions are covered too.
- **sdk**: `/cancel` frees the slot instead of only setting the flag, so a hung task can't block a source until the process restarts.
- **connector-manager**: cancel writes the DB state before telling the connector, so a run is never `running` here and slot-free there.
- **connector-manager**: the breaker reopens after the maximum backoff instead of staying shut. A successful run clears the streak; a failed one starts the wait again. No new config — it reuses `SYNC_BACKOFF_MAX_SECONDS`.

Any one of these breaks the loop; together they make each step recover on its own.

## Tests

`cargo test` for `omni-connector-manager` (75), `omni-imap-connector` (14) and the SDK's `server_integration_test` (18), all green. Adds one scheduler case for the breaker reopening; it fails on master and passes here.


prvnsmpth:
Thanks for the fix! @ramteid

